PR #22, opened 2018-11-30
Library linking

rzadp:
Added `link` to the waffle lib.
Added usage to Readme.


marekkirejczyk:
Published in Waffle@1.0.5

PR #25, opened 2018-12-06
Support for faster compilation with solc

marekkirejczyk:


snario:
Hey dude heads up you're publishing to `npm` on this branch (https://www.npmjs.com/package/ethereum-waffle/v/1.1.2).

Broke my build :p

marekkirejczyk:
@snario That supposed to be an experimental version, but without breaking changes. Why would it break your build? Use ^1.0.5 for now.

snario:
@marekkirejczyk yeah it turned out that the problem was #26. My bad, it was 3 AM :p

Issue #26, opened 2018-12-06
Solidity bigNumber checker breaks if ethers versions are not the same


snario:
This is because the matcher uses `instanceOf`

snario:
`ethers.utils.BigNumber` has a `.isBigNumber` method

PR #27, opened 2018-12-06
Replace instanceOf with BigNumber.isBigNumber

snario:
Fixes https://github.com/EthWorks/Waffle/issues/26


marekkirejczyk:
Published in Waffle@1.05

PR #37, opened 2019-01-09
Fix import directory mapping for inside out directory structure

marekkirejczyk:


marekkirejczyk:
Already merged in #38

PR #41, opened 2019-01-11
Typescript rewrite

sz-piotr:
This PR rewrites the Waffle source code in typescript with minimal changes. It also swaps the eslint linter for tslint.

No type declarations are emitted in the build output as of this PR. The reasoning is that some apis need more care given to them before typings can be provided.

Some of the tests are still left in javascript (namely matcher tests). This is because in order to provide type information for chai, specialised declaration files are required similar to [chai-as-promised](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/chai-as-promised/index.d.ts). Those type declarations cannot be inferred from the typescript source.

Additionally shim type information had to be created for the `solc` library.

The compiler settings in this PR are also quite permissive and need to be tightened in the future. The only strict option is `noImplicitAny`. Using `strict: true` in the future is highly recommended.

This PR builds upon https://github.com/EthWorks/Waffle/pull/39 and contains typescript version of the fixtures.


marekkirejczyk:
Seems like a good first step to full TypeScript Migration. ❤️ ❤️ ❤️ ❤️

PR #62, opened 2019-01-22
Use rimraf in clean script instead of rm -rf

yaram:
The command `rm -rf` only works on unix-likes, thus the clean script fails on Windows. This replaces the command with the `rimraf` command from the package of the same name. It is a CLI tool that works identically to the `rm -rf` command, removing a file or an entire directory, but is implemented in platform-independent JS.


marekkirejczyk:
Feature is now released in [2.0.1](https://github.com/EthWorks/Waffle/releases/tag/2.0.1)

Thanks for great work!

PR #72, opened 2019-01-29
Added support for standardised import resolution

ritave:
[Resolver-Engine](https://github.com/Crypto-Punkers/resolver-engine) project is an [Ethereum Foundation grantee](https://blog.ethereum.org/2018/08/17/ethereum-foundation-grants-update-wave-3/), with a goal to create tooling for standardised Solidity import resolution.

Currently each and every framework has a different way of resolving imports, causing great anguish for the end-developers. With the Resolver-Engine, all projects that use it have a common set of supported functionality, sane defaults and maintenance of a dedicated team meaning that in the future Waffle will be able to support more imports just by upgrading the dependency.

Basic usage:
```typescript
import { ImportFsEngine } from "@resolver-engine/import-fs";

ImportFsEngine()
  .require("github:OpenZeppelin/openzeppelin-solidity/contracts/ownership/Ownable.sol#v2.1.2")
  .then(file => console.log(`Loaded url ${file.url} with source: ${file.source}`);
```

Existing functionality that instantly gets added to this project:
- [x] Filesystem, relative and absolute imports
- [x] NPM and node_modules with support for monorepo pattern
- [x] URI and URL support
- [x] Github and branch support
- [x] IPFS through a gateway
- [x] Swarm through a gateway
- [x] Initial Truffle-like support for Ethpm packages

Additionally, backwards compatibility was added to support Waffle's original resolution methods

**We've noticed a refactor in the past days, thus we mark this PR as a Work-In-Progress**, so that we can work with you, gather feedback and together create a [Cooperative](https://en.wikipedia.org/wiki/Cooperative) community around the tooling and standards.

In the upcoming days similiar PRs will be posted to
- Remix-IDE
- 0xProject sol-compiler


marekkirejczyk:
Hi @ritave, thank you for great work!

I think it is a great idea. We would super happy to (be the first to?) use it!

I would propose the following next steps:
1) Split PR into two: 
 - changes related to ResolverEngine
 - all styling changes move to separate PR (including) .prettierrc.yaml as they generate some noise in the repository and make it hard to analyze (preferably remove leading and trailing spaces for `{`, `{`)

2) Update PR to resolve conflicts. 
While there was a significant amount of changes, the core concepts didn't change, so it should be fairly easy. 

Very excited to merge it! Looking forward! ❤️ ❤️ ❤️ ❤️

ritave:
@marekkirejczyk Thanks for the feedback, I've cleaned up the change and made it work after the major refactor. It's now ready to be merged in!

lib/compiler/buildUitls.ts
```diff
@@ -1,3 +1,11 @@
+import { ImportFile } from '@resolver-engine/imports';
+
+export function buildSourcesObject(files: ImportFile[]): Record<string, { content: string}> {
```
sz-piotr: This function could actually be called by `buildInputObject` itself. This would simplify the calling code. Previously it had to be separate because they were differences between the compilers

lib/compiler/compileDocker.ts
```diff
@@ -1,16 +1,16 @@
...
 
 const CONTAINER_PATH = '/home/project';
 const NPM_PATH = '/home/npm';
 
 export function compileDocker(config: Config) {
-  return async function compile(sources: string[]) {
+  return async function compile(sources: ImportFile[]) {
     const command = createBuildCommand(config);
-    const input = JSON.stringify(buildInputJson(sources, config), null, 2);
+    const input = JSON.stringify(buildInputObject(buildSourcesObject(sources)), null, 2);
```
sz-piotr: This could be simplified to just `buildInputObject(sources)` by having `buildInputObject` call `buildSourcesObject`

lib/compiler/compileNative.ts
```diff
@@ -1,13 +1,13 @@
...
-import {buildInputObject, buildSources} from './buildUitls';
+import {buildInputObject, buildSourcesObject} from './buildUitls';
+import { ImportFile } from '@resolver-engine/imports';
 
 export function compileNative(config: Config) {
-  return async function compile(sources: string[]) {
+  return async function compile(sources: ImportFile[]) {
     const command = createBuildCommand(config);
-    const input = JSON.stringify(buildInputJson(sources, config), null, 2);
+    const input = JSON.stringify(buildInputObject(buildSourcesObject(sources)), null, 2);
```
sz-piotr: This could be simplified to just `buildInputObject(sources)` by having `buildInputObject` call `buildSourcesObject`

lib/compiler/compileSolcjs.ts
```diff
@@ -15,21 +16,14 @@ async function loadCompiler(config: Config) {
 }
 
 export function compileSolcjs(config: Config) {
-  return async function compile(sources: string[], findImports: (file: string) => any) {
+  return async function compile(sources: ImportFile[], findImports: (file: string) => any) {
     const solc = await loadCompiler(config);
-    const inputs = findInputs(sources);
-    const input = buildInputObject(convertInputs(inputs));
+    const input = buildInputObject(buildSourcesObject(sources));
```
sz-piotr: This could be simplified to just `buildInputObject(sources)` by having `buildInputObject` call `buildSourcesObject`

lib/compiler/compiler.ts
```diff
@@ -16,9 +18,14 @@ export async function compileAndSave(config: Config) {
 }
 
 export async function compile(config: Config) {
+  // Added support for backwards compatibillity - renamable node_modules path
+  const resolver = ImportsFsEngine()
+    .addResolver(resolvers.BacktrackFsResolver(config.npmPath));
+  const cache = await gatherSourcesAndCanonizeImports(findInputs(config.sourcesPath), process.cwd(), resolver);
```
sz-piotr: This line is quite long, having each argument in its own line would definitely increase readability

test/example/example.ts
```diff
@@ -50,7 +51,10 @@ describe('INTEGRATION: Example', () => {
 
   it('should use library to add 7', async () => {
     const myLibrary = await deployContract(wallet, MyLibrary, []);
-    link(LibraryConsumer, 'test/projects/example/MyLibrary.sol:MyLibrary', myLibrary.address);
+    link(
+      LibraryConsumer,
+      `${path.join(process.cwd(), 'test/projects/example/MyLibrary.sol')}:MyLibrary`,
+      myLibrary.address);
```
sz-piotr: Add a newline here, before `)`

lib/compiler/compiler.ts
sz-piotr: I would suggest renaming `cache` to `sources` to match the rest of the project

lib/compiler/findImports.ts
```diff
@@ -1,20 +1,11 @@
-import fs from 'fs';
-import path from 'path';
-import {readFileContent} from '../utils';
+import { ImportFile } from '@resolver-engine/imports';
 
-export function findImports(libraryPath: string) {
+export function findImports(fileCache: ImportFile[]) {
```
sz-piotr: I would suggest renaming `fileCache` to `sources` to match the rest of the project

test/compiler/e2e.ts
```diff
@@ -107,7 +107,7 @@ describe('E2E: Compiler integration', () => {
         const MyLibrary = require(libraryPath);
         const LibraryConsumer = deepCopy(require(resolve(join(configuration.targetPath, 'Two.json'))));
         const myLibrary = await deployContract(wallet, MyLibrary, []);
-        const libraryName = `${configuration.sourcesPath.slice(2)}/MyLibrary.sol:MyLibrary`;
+        const libraryName = `${join(process.cwd(), configuration.sourcesPath.slice(2))}/MyLibrary.sol:MyLibrary`;
```
marekkirejczyk: #breakingchange This is possibly breaking change. That would imply, that linking a library now requires an absolute path, which would be most of the time impossible to do, as people link libraries on different machines.

test/example/example.ts
marekkirejczyk: #breakingchange

test/compiler/compiler.ts
```diff
@@ -25,7 +30,7 @@ describe('INTEGRATION: Compiler', () => {
     const sourcesPath = './test/projects/invalidContracts'; // tslint:disable-line
     const targetPath = './test/projects/build'; // tslint:disable-line
     const config = {...defaultConfig, sourcesPath, targetPath}; // tslint:disable-line
-    const expectedOutput = 'test/projects/invalidContracts/invalid.sol:6:14: ' +
+    const expectedOutput = `${path.join(process.cwd(), 'test/projects/invalidContracts/invalid.sol')}:6:14: ` +
```
marekkirejczyk: #breakingchange

test/compiler/compiler.ts
```diff
@@ -15,7 +16,11 @@ chai.use(sinonChai);
 describe('INTEGRATION: Compiler', () => {
   it('compile just compiles', async () => {
     const output = await compile(config);
-    const basicTokenOutput = output.contracts['test/projects/example/BasicToken.sol'].BasicToken;
+    const basicTokenOutput = output.contracts[
+      path.join(
+        process.cwd(),
```
marekkirejczyk: #breakingchange

test/compiler/compiler.ts
marekkirejczyk: #breakingchange

marekkirejczyk:
Hi @ritave,

I did re-review and found potentially breaking change -> linking now requires a full path of a file, which is impossible to use in practice (as people use different computers, with different file systems).

That might be a symptom of a deeper problem, that Resolver-Engine is using file paths instead of fully qualified names.

marekkirejczyk:
Ping @ritave, any news? :D

squadack:
Hi,
Thanks for your feedback. We've managed to address all your remarks and maintain compatibility with previous feature set - linking now works the same way as before.

test/example/example.ts
```diff
@@ -50,7 +51,11 @@ describe('INTEGRATION: Example', () => {
 
   it('should use library to add 7', async () => {
     const myLibrary = await deployContract(wallet, MyLibrary, []);
-    link(LibraryConsumer, 'test/projects/example/MyLibrary.sol:MyLibrary', myLibrary.address);
+    link(
+      LibraryConsumer,
+      'test/projects/example/MyLibrary.sol:MyLibrary',
+      myLibrary.address
+    );
```
marekkirejczyk: 👍

marekkirejczyk:
Looks good, nice refactoring! Merging. Will release a new version soon.

PR #99, opened 2019-04-03
Fix #97 extending to async user configuration file

### LINKED ISSUE #97: Feature request: loadConfig as a promise
Pain point:
I'm trying to assign different waffle config in a `waffle.js` on the fly based on the build environment, counterfactual/monorepo#1211
Yet it would be an async operation, thus just by running `waffle waffle.js` would only gives me default settings.

What it look like in my `waffle.js`:
```js
module.exports = new Promise((resolve, reject) => {
  getConfig((result) => {
    resolve(result);
  });
});
```

AlexXiong97:


AlexXiong97:
@marekkirejczyk this is PR is ready for review and merge

AlexXiong97:
counterfactual/monorepo#1211 should be aware of this.

lib/config/loadConfig.ts
```diff
@@ -8,9 +8,14 @@ function readConfigFile(configPath: string) {
   return {};
 }
 
-export function loadConfig(configPath: string): Config {
-  return {
-    ...defaultConfig,
-    ...readConfigFile(configPath)
-  };
+export async function loadConfig(configPath: string): Promise<object> {
```
marekkirejczyk: Wy not `Promise<Config>`?

test/compiler/e2e.ts
```diff
@@ -46,7 +48,11 @@ describe('E2E: Compiler integration', () => {
   });
 
   for (const configurationPath of configurations)  {
-    const configuration = loadConfig(configurationPath) as any;
+    const userConfig = await loadConfig(configurationPath);
+    const configuration = {
```
marekkirejczyk: I would propose to extract it to: 
```ts
const getConfig(configurationPath) : Configuration {
    const userConfig = await loadConfig(configurationPath);
    return {
      ...defaultConfig,
      ...userConfig
    };
}
```

Usage:
```js
const {name, targetPath, legacyOutput} = getConfig(configurationPath);
```

lib/compiler/compiler.ts
```diff
@@ -9,7 +9,11 @@ import {ImportsFsEngine, resolvers} from '@resolver-engine/imports-fs';
 import {gatherSources} from '@resolver-engine/imports';
 
 export async function compileProject(configPath: string) {
-  await compileAndSave(loadConfig(configPath));
+  const userConfig = await loadConfig(configPath);
+  await compileAndSave({
+    ...defaultConfig,
+    ...userConfig
+  });
```
marekkirejczyk: I would propose to extract it to: 
```ts
const getConfig = async (configurationPath) : Promise<Configuration> {
    const userConfig = await loadConfig(configurationPath);
    return {
      ...defaultConfig,
      ...userConfig
    };
}
```

Usage:
```js
await compileAndSave(await getConfig(configPath));
```

marekkirejczyk:
This is a great idea. Good work. 👍 👍 👍 
I am happy to merge it.

I would propose two improvements:
- some cosmetic improvements in the comments
- add a short note about it in the documentation

lib/config/loadConfig.ts
AlexXiong97: Originally, because the if the user option is empty, then return an empty object, now I will change to returning a `defaultConfig` and take your comment. thanks

test/compiler/e2e.ts
AlexXiong97: points taken, but since the variable `configuration` will be used at https://github.com/EthWorks/Waffle/blob/ee1b9d54e807817f58aa511a95ff8a716e66653e/test/compiler/e2e.ts#L112
will simplify my last commit but keep configuration as a const var

AlexXiong97:
@marekkirejczyk ready for review :rocket:

marekkirejczyk:
Released in [@2.0.9](https://github.com/EthWorks/Waffle/releases/tag/2.0.9)

Issue #92, opened 2019-03-28
Import from Node Modules does not work on Windows


tchataigner:
Hello all !

I just saw this package and I wanted to try it out for writing and testing some contracts. 

Because of work reglementations I am forced to develop on a Windows 10 OS. Everything went well until I tried to import the ERC721 from the Open Zepplin Package. 

`import "openzeppelin-solidity/contracts/token/ERC721/ERC721Full.sol";`

When doing so I got this error : 

`Error: Invalid URI "c:/Users/tchataignerprestatai/WebstormProjects/artikodin-contracts/node_modules/openzeppelin-solidity/contracts/token/ERC721/ERC721Full.sol"`

After looking from where it might come from I found that this test is going on in the package 
`if (!(self.uri.host || (self.uri.hostname && self.uri.port)) && !self.uri.isUnix)` (from request package, request.js line 260)
and its where the code is returning.

Any idea of how to solve that ?

yaram:
I've identified the issue as being in the `@resolver-engine/core` package. I've submitted an issue here Crypto-Punkers/resolver-engine#182

yaram:
This issue has been fixed in Crypto-Punkers/resolver-engine#184

PR #111, opened 2019-07-05
Update resolver-engine to 0.3.3 to fix issue with windows paths

yaram:
Update the `@resolver-engine/imports` and `@resolver-engine/imports-fs` dependencies to version `^0.3.3`, fixing #92.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
     "node": ">=10.0"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


yaram:
It's worth mentioning that that the `compiler/findInputs` tests do not pass under Windows, as they check for equality against a unix-like path.

sz-piotr:
Thanks for the contribution 👍

PR #113, opened 2019-07-11
Update dependencies

sz-piotr:


snario:
This fixed `solc` to `0.5.1` which is no good for us @marekkirejczyk @sz-piotr

sz-piotr:
@snario We run into runtime error issue using versions >0.5.1. Until they we resolve them saying that we support them is lying. We are currently working on a fix

marekkirejczyk:
@snario New version released. Should be working now. Feedback welcome.

PR #115, opened 2019-07-17
Overload deployContract

sz-piotr:


marekkirejczyk:
@sz-piotr  What about documentation?

Issue #126, opened 2019-08-14
Error: Cannot find module 'uuid/v4'


gtklocker:
Repo with README and steps to reproduce:

https://github.com/gtklocker/waffle-bug-report

marekkirejczyk:
Hi @gtklocker,

Thank you for reporting. 

Fixed in [#131](https://github.com/EthWorks/Waffle/pull/131)
Released in [2.1.0](https://github.com/EthWorks/Waffle/releases/tag/2.1.0)

Closing. Please reopen if still an issue.

rad-p:
Still have the same 
`❯ yarn test               
yarn run v1.19.1
$ waffle && mocha
Error: Cannot find module 'uuid/v4'
Require stack:
- .node_modules/ganache-core/build/ganache.core.node.js
- .node_modules/ganache-core/index.js
- .node_modules/ethereum-waffle/dist/waffle.js
- .test/basicTokenMock.js
- .node_modules/mocha/lib/mocha.js
- .node_modules/mocha/lib/cli/one-and-dones.js
- .node_modules/mocha/lib/cli/options.js
- .node_modules/mocha/bin/mocha
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:794:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:849:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (.node_modules/ganache-core/build/webpack:/Ganache/external "safe-buffer":1:1)
    at __webpack_require__ (.node_modules/ganache-core/build/webpack:/Ganache/webpack/bootstrap:9:1)
    at Object.<anonymous> (.node_modules/ganache-core/build/webpack:/Ganache/node_modules/ethereumjs-wallet/index.js:3:141)
    at __webpack_require__ (.node_modules/ganache-core/build/webpack:/Ganache/webpack/bootstrap:9:1)
    at Object.GethApiDouble (.node_modules/ganache-core/build/webpack:/Ganache/lib/subproviders/geth_api_double.js:850:1)
    at __webpack_require__ (.node_modules/ganache-core/build/webpack:/Ganache/webpack/bootstrap:9:1) 
    at Object.from (.node_modules/ganache-core/build/webpack:/Ganache/lib/subproviders/gethdefaults.js:32:1)
[...]

PR #131, opened 2019-08-23
Fix ganache-core version, yarn.lock update

cehali:
There is new ganache-core which unfortunately is buggy and this is why I fixed version 2.6.1. Also yarn.lock hasn't been updated for a long time, so I update it to use the same packages' versions during development as users when they add Waffle as a dependency.


cehali:
Also, PR fixes the issue: https://github.com/EthWorks/Waffle/issues/126

PR #135, opened 2019-08-27
remove always-true interface key check

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
solidity doesn't generate `interface`, so this check always succeeds


sz-piotr:
Unfortunately now people might rely on this behaviour, so this is actually a breaking change. We are planning to address this as part of Waffle 3. See #155

PR #137, opened 2019-08-30
Integrate ENSBuilder

cehali:


lib/ENSBuilder/ENSBuilder.ts
```diff
@@ -0,0 +1,95 @@
...
+import ReverseRegistrar from './abi/ReverseRegistrar.json';
+import {Registrars} from './ENSBuilderTypes';
+
+const overrideOptions = {gasLimit: 120000};
+const nullLogger = () => {};
+
+class ENSBuilder {
+  public deployer: Wallet;
+  public registrars: Registrars = {};
+  public logger: any;
```
Marik-D: Why not set a stricter type for it: `(str: string) => void`?

lib/ENSBuilder/ENSBuilderTypes.ts
```diff
@@ -0,0 +1,5 @@
+import {Contract} from 'ethers';
+
+export interface Registrars {
```
Marik-D: Can be `export type Registrars = Record<string, Contract>`

marekkirejczyk:
Will continue in new PR

PR #147, opened 2019-10-22
Buidler Integration

Anze1m:


lib/compiler/compileSolcjs.ts
```diff
@@ -3,7 +3,7 @@ import {promisify} from 'util';
 import {readFileContent, isDirectory, relativePathToWorkingDir} from '../utils';
 import {Config} from '../config/config';
 import {buildInputObject} from './buildUitls';
-import { ImportFile } from '@resolver-engine/imports';
+import {ImportFile} from '@resolver-engine/imports';
```
marekkirejczyk: Good stuff on fixing formatting. In the future, this could go as a separate PR.

lib/fixtures.ts
```diff
@@ -22,7 +22,7 @@ export function createFixtureLoader(overrideProvider?: providers.Web3Provider, o
       snapshot.id = await snapshot.provider.send('evm_snapshot', []);
       return snapshot.data;
     } else {
-      const provider = overrideProvider || createMockProvider();
+      const provider = overrideProvider || createGanacheProvider();
```
marekkirejczyk: We can't break backward compatibility. I would propose to have 3 functions:
```ts
createGanacheProvider() - creates ganache provider
createBuidlerProvider() - create Buidler provider
createMockProvider() - creates default provider (ganache by default, can be overridden in config)
```

lib/provider.ts
```diff
@@ -1,22 +1,55 @@
 import Ganache, {GanacheOpts} from 'ganache-core';
+import {createProvider} from '@nomiclabs/buidler/internal/core/providers/construction';
```
marekkirejczyk: Split into 3 files? 

```
provider.ts (createMockProvider)
ganacheProvider.ts
buidlerProvider.ts
```

package.json
```diff
@@ -24,6 +24,8 @@
     "node": ">=10.0"
   },
   "dependencies": {
+    "@nomiclabs/buidler": "^1.0.1",
+    "@nomiclabs/buidler-ethers": "^1.0.1",
```
marekkirejczyk: Shall we add them as peer dependencies to reduce the amount of dependencies?

lib/provider.ts
```diff
@@ -1,22 +1,55 @@
...
+  blockGasLimit: 8000000000000,
+  hardfork: 'petersburg',
+  throwOnTransactionFailures: true,
+  throwOnCallFailures: true,
+  chainId: 17
+  };
 
-export function createMockProvider(ganacheOptionsOrPathToConfig: string | GanacheOpts = {}) {
+export const createMockProvider = (providerOptionsOrPathToConfig: string | GanacheOpts = {}) => {
+  console.warn(`WARNING! 'createMockProvider' method is deprecated. Use 'createGanacheProvider' or 'createBuidlerProvider' instead.`);
```
marekkirejczyk: no need to deprecate (yet!)

test/compiler/e2e.ts
```diff
@@ -113,7 +113,7 @@ describe('E2E: Compiler integration', async () => {
       });
 
       it('links library', async () => {
-        const provider = createMockProvider();
+        const provider = createGanacheProvider();
```
marekkirejczyk: #providerDefaultMethod

test/deployContract.ts
```diff
@@ -1,11 +1,10 @@
 import {expect} from 'chai';
 import {ContractFactory} from 'ethers';
-import {deployContract, createMockProvider, getWallets} from '../lib';
-import {ContractJSON} from '../lib/ContractJSON';
+import {deployContract, createGanacheProvider, getWallets, ContractJSON} from '../lib';
 import SafeMath from './example/build/SafeMath.json';
 
 describe('deployContract', () => {
-  const provider = createMockProvider();
+  const provider = createGanacheProvider();
```
marekkirejczyk: #providerDefaultMethod

test/example/example.ts
```diff
@@ -1,52 +1,69 @@
...
 
-  beforeEach(async () => {
-    token = await deployContract(wallet, BasicTokenMock, [wallet.address, 1000]);
-  });
+  let provider: providers.JsonRpcProvider;
 
-  it('Assigns initial balance', async () => {
-    expect(await token.balanceOf(wallet.address)).to.eq(1000);
-  });
+  providerCreators.forEach((providerCreator) => {
```
marekkirejczyk: Me Gusta 👍

test/matchers/reverted.ts
```diff
@@ -1,10 +1,10 @@
 import {expect} from 'chai';
-import {createMockProvider, deployContract, getWallets} from '../../lib';
+import {createGanacheProvider, getWallets, deployContract} from '../../lib';
 import Matchers from './build/Matchers.json';
-import { Contract } from 'ethers';
+import {Contract} from 'ethers';
 
 describe('INTEGRATION: Matchers: reverted', () => {
-  const provider = createMockProvider();
+  const provider = createGanacheProvider();
```
marekkirejczyk: #providerDefaultMethod

lib/fixtures.ts
Anze1m: This is more or less how it works, but with one difference. To keep typing safe, `createMockProvider ` returns the same what `createGanacheProvider ` returns, with no additional options (but additionally warns that method is deprecated). In this particular line, we want to create Ganache provider, because Buidler does not support snapshots and reverts.

marekkirejczyk:
Documentation is missing. We should include:
- info that now we have two providers (and one is the default)
- the new provider is experimental for now
- how to use the second provider (add peer dependency)
- limitation of the new provider (no support for fixtures)
- release notes

lib/fixtures.ts
Anze1m: Ok, if we don't deprecate `createMockProvider` yet, then we can keep using it in the code.

test/example/example.ts
marekkirejczyk: We should extract providers test setup to a separate file and use both providers in all tests related to providers, not only example (except fixtures ofc)

lib/compiler/compileSolcjs.ts
Anze1m: https://github.com/EthWorks/Waffle/pull/148 <- future is now :sunglasses:

package.json
Anze1m: We can, but these two buidler packages are far smaller than `ganache-core`. Is there really need for that?

lib/compiler/compileSolcjs.ts
Anze1m: Separate PR is a good idea, but I'd rather prefer to merge it before this one.

Anze1m: 👌

Anze1m:
Maybe we want to introduce `forEachProvider` not only as an internal helper but as a fully-fledged feature? What do you think @marekkirejczyk?

lib/buidlerProvider.ts
```diff
@@ -0,0 +1,30 @@
...
+  accounts: defaultAccounts.map((account) => ({
+    balance: account.balance,
+    privateKey: account.secretKey
+  })),
+  blockGasLimit: 8000000000000,
+  hardfork: 'istanbul',
+  throwOnTransactionFailures: true,
+  throwOnCallFailures: true,
+  chainId: 17
+  };
```
Anze1m: Any opinions on the defaults? Especially `blockGasLimit`, `hardfork` and `chainId `.

PR #150, opened 2019-11-01
Fix #149: assertion error message for non-existent events

### LINKED ISSUE #149: Improve assertion error message when event does not exist
I have a simple contract:

```solidity
contract Counter {
  uint count = 0;

  event CountedTo(uint number);
  // Remaining omitted for brevity
```
Then in my tests, I have:

```ts
    await expect(counter.countDown())
      .to.emit(counter, "CountedTo")
      .withArgs(10);
```

This works as expected.

However, at first I misspelled the event name and wrote the assertion like this:

```ts
    await expect(counter.countDown())
      .to.emit(counter, "CountTo") // <---------- HERE
      .withArgs(10);
```

Then, instead of getting a nice error saying such event does not exist, what I got was:

```
TypeError: Cannot read property 'topic' of undefined
      at <dir>/node_modules/ethereum-waffle/dist/matchers/matchers.js:63:62
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

But if such `CountTo` event does exist in the contract, but wasn't emitted during the call, it displays the following error message:

```
AssertionError: Expected event "CountTo" to emitted, but wasn't
      + expected - actual

      +CountTo
```

So, it's working as expected.

Ideally we could have a message warning the user when such event isn't even declared in the contract it would be nice.

I'll try to find where in the code base I could patch this behavior and try to come up with something.

hbarcelos:
- [x] Check for non-existent events on `emit` assertion helper to provide proper feedback.

### DIFF
--- a/lib/matchers/matchers.ts
+++ b/lib/matchers/matchers.ts
@@ -80,7 +80,26 @@ const solidity = (chai: any, utils: any) => {
     const derivedPromise = promise.then((tx: any) =>
       contract.provider.getTransactionReceipt(tx.hash)
     ).then((receipt: any) => {
-      const {topic} = contract.interface.events[eventName];
+      const eventDescription = contract.interface.events[eventName];
+
+      if (eventDescription === undefined) {
+        const isNegated = this.__flags.negate === true;
+
+        this.assert(
+          isNegated,
+          `Expected event "${eventName}" to be emitted, but it doesn't` +
+          ` exist in the contract. Please make sure you've compiled` +
+          ` its latest version before running the test.`,
+          `WARNING: Expected event "${eventName}" NOT to be emitted.` +
+          ` The event wasn't emitted because it doesn't` +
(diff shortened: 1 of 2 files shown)


lib/matchers/matchers.ts
```diff
@@ -80,7 +80,26 @@ const solidity = (chai: any, utils: any) => {
     const derivedPromise = promise.then((tx: any) =>
       contract.provider.getTransactionReceipt(tx.hash)
     ).then((receipt: any) => {
-      const {topic} = contract.interface.events[eventName];
+      const eventDescription = contract.interface.events[eventName];
+
+      if (eventDescription === undefined) {
+        const isNegated = this.__flags.negate === true;
```
hbarcelos: I'm not sure this is the way to go when extending `chai` with a plugin. Couldn't find much information in the docs.

sz-piotr:
Merged as part of #151

PR #151, opened 2019-11-01
Fix #117: emit method name on docs

### LINKED ISSUE #117: Update `expect(...).to.emit(...)` id docs
The matcher to check events emitted from transaction is called `emit`. While in docs it is still written as `emitEvent`.

https://github.com/EthWorks/Waffle/blame/master/README.md#L18

hbarcelos:
Update README to reflect API change for `chai` solidity matchers:
- `emitEvent` -> `emit`


sz-piotr:
Thanks for this PR ❤️

PR #159, opened 2019-12-05
quick fix to handle BigNumbers array as argument of the event log

bytniak:
expected:
as        ```[ BigNumber { _hex: '0x01f4' }, BigNumber { _hex: '0x01f4' } ]``` 
equals ```[ BigNumber { _hex: '0x01f4' }, BigNumber { _hex: '0x01f4' } ]``` 
- it should passes the test

as        ```[ BigNumber { _hex: '0x01f4' }, BigNumber { _hex: '0x01f4' } ]``` 
equals ```[ 500, 500 ]``` 
- it should passes the test

actual:
```
AssertionError: expected [ Array(2) ] to equal [ Array(2) ]
    at assertArgsArraysEqual (src/test/productAssetIdentity.spec.ts:65:51)
    at derivedPromise.promise.then (src/test/productAssetIdentity.spec.ts:71:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
```
AssertionError: expected [ Array(2) ] to equal [ 500, 500 ]
    at assertArgsArraysEqual (src/test/productAssetIdentity.spec.ts:65:51)
    at derivedPromise.promise.then (src/test/productAssetIdentity.spec.ts:71:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
switching to deep.equal also doesn't solve the issue...

quick and dirty fix is to stringify the `actualArgs` and `expectedArgs` values of every argument within `withArgs` method.

however this makes `equal` method overwrites to handle BigNumbers useless...

as an alternative if the value is an array we can compare each value separately (last commit)

### DIFF
--- a/lib/matchers/matchers.ts
+++ b/lib/matchers/matchers.ts
@@ -122,7 +122,7 @@ const solidity = (chai: any, utils: any) => {
       expectedArgs.length,
       actualArgs.length);
     for (let index = 0; index < expectedArgs.length; index++) {
-      new chai.Assertion(expectedArgs[index]).equal(actualArgs[index]);
+        new chai.Assertion(JSON.stringify(expectedArgs[index])).equal(JSON.stringify(actualArgs[index]));
     }
   };
 
--- a/lib/matchers/matchers.ts
+++ b/lib/matchers/matchers.ts
@@ -121,15 +121,15 @@ const solidity = (chai: any, utils: any) => {
       'Do not combine .not. with .withArgs()',
       expectedArgs.length,
       actualArgs.length);
-      for (let index = 0; index < expectedArgs.length; index++) {
-        if (expectedArgs[index].length !== undefined && typeof expectedArgs[index] !== "string"){
-            for (let j = 0; j < expectedArgs[index].length; j++){
-                new chai.Assertion(expectedArgs[index][j]).equal(actualArgs[index][j]);
-            }
-        } else {
-            new chai.Assertion((expectedArgs[index])).equal((actualArgs[index]));
+    for (let index = 0; index < expectedArgs.length; index++) {
+      if (expectedArgs[index].length !== undefined && typeof expectedArgs[index] !== 'string') {
+        for (let j = 0; j < expectedArgs[index].length; j++) {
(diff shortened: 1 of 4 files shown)


marekkirejczyk:
Are you using `expect(arr1).to.deep.eq(arr2)`?

bytniak:
The problem is within `.to.emit().withArgs()` method and precisely `assertArgsArraysEqual` helper function. 

(In `assertArgsArraysEqual`)
> switching to deep.equal also doesn't solve the issue...

The issue is when you have as one of the event arguments a big number array. I updated the initial comment for clarity.

lib/matchers/matchers.ts
```diff
@@ -121,9 +121,15 @@ const solidity = (chai: any, utils: any) => {
...
-    }
+      for (let index = 0; index < expectedArgs.length; index++) {
+        if (expectedArgs[index].length !== undefined && typeof expectedArgs[index] !== "string"){
+            for (let j = 0; j < expectedArgs[index].length; j++){
+                new chai.Assertion(expectedArgs[index][j]).equal(actualArgs[index][j]);
+            }
+        } else {
+            new chai.Assertion((expectedArgs[index])).equal((actualArgs[index]));
+        }
+     }
```
marekkirejczyk: LGTM
Would you:

- add a test [here](https://github.com/EthWorks/Waffle/blob/master/test/matchers/events.ts) (will need to add example event [here](https://github.com/EthWorks/Waffle/blob/master/test/matchers/contracts/events.sol))
- perhaps extract method `assertArrayEqual(expectedArgs, actualArgs)`

Happy to merge and release after that.

bytniak: added tests and problematic event
passes eslint tests 
didn't extract the `assertArrayEqual` method for now

marekkirejczyk:
@bytniak Awesome work! 👍👍👍 
Thank you! ❤️❤️❤️
 
Published in [2.1.2](https://github.com/EthWorks/Waffle/releases/tag/2.1.2)

bytniak:
Thx! @marekkirejczyk.

PR #178, opened 2019-12-28
Split waffle into multiple packages

sz-piotr:
# PLEASE READ THIS

This PR does two things. The split and small api changes for providers. Why does it do the api changes? Because it simplifies the code and adds more type safety. There are no breaking changes, only deprecations.

**MERGING THIS PR WILL REQUIRE DEPLOYING ADDITIONAL PACKAGES!**

Now `ethereum-waffle` will depend on `@waffle/compiler`, `@waffle/provider` and `@waffle/chai`. An npm org of `@waffle` has to be created and the appropriate packages published.

## Packages
- extract matchers to `@waffle/chai`
- extract provider creation and fixtures to `@waffle/provider`
- extract compilation to `@waffle/compiler`
- move waffle code to `/waffle-waffle` (keeps package name of `ethereum-waffle`)
## Api surface
- **deprecate** `createMockProvider`
- **deprecate** `getGanacheOptions` - **note**: was anyone using this directly? Will anyone miss `createMockProvider(path)`?
- **deprecate** `getWallets`
- introduce `MockProvider` with `getWallets` and `createEmptyWallet`
- fixtures now enforce the type of `MockProvider`. The didn't really work with anything else so that is not a breaking change.


marekkirejczyk:
@waffle is already taken: 
https://www.npmjs.com/org/waffle

.circleci/config.yml
```diff
@@ -1,79 +1,65 @@
...
+jobs:
+  lint_and_build:
+    docker:
+      - image: circleci/node:11.15
+    steps:
+      - checkout
+      - install_dependencies:
+          key: 11
+      - run: yarn lint
+      - run: yarn build
```
marekkirejczyk: We are already building in with node 11 with `test_node_11` job, why duplicate?

.gitignore
```diff
@@ -1,15 +1,5 @@
...
-/test/compiler/build/
-/test/projects/custom/custom_build
-/test/projects/insideOut/build
-/test/projects/custom_solidity_4/custom_build
-/test/projects/compilerOptions/build
-/test/projects/humanReadableAbi/build
-/test/matchers/build/
-/test/example/build/
-/build
-/tmp
```
marekkirejczyk: `.gitignore` for `waffe-provider` and `waffle-chai` seems to be missing

waffle-compiler/.npmignore
```diff
@@ -0,0 +1,5 @@
+src
```
marekkirejczyk: `.gitignore` seems to be missing

waffle-chai/.npmignore
```diff
@@ -0,0 +1,5 @@
+src
```
marekkirejczyk: .gitignore missing

waffle-waffle/LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright 2019 Marek Kirejczyk
```
marekkirejczyk: Hmmm... We could change to Ethworks... or list key contribiutors

waffle-chai/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@waffle/chai",
+  "description": "A sweet set of chai matchers for your blockchain testing needs.",
```
marekkirejczyk: 👍

package.json
```diff
@@ -1,67 +1,18 @@
...
-  },
-  "resolutions": {
-    "web3": "1.2.1"
+  "workspaces": [
+    "waffle-*"
+  ],
+  "scripts": {
+    "lint": "wsrun -c lint",
+    "build": "wsrun -te -c build",
+    "test": "wsrun -c test"
```
marekkirejczyk: Would be nice to have "publish" here as well.

waffle-chai/test/index.ts
```diff
@@ -0,0 +1,6 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import {waffleChai} from '../src';
+
+chai.use(chaiAsPromised);
+chai.use(waffleChai);
```
marekkirejczyk: BEFORE:
```ts
import {solidity} from 'waffle';
chai.use(solidity);
```

NOW:
```ts
import {waffleChai} from '@waffle/waffle-chai';
chai.use(waffleChai);
```

waffle-chai/test/contracts/Events.ts
```diff
@@ -0,0 +1,53 @@
+export const EVENTS_SOURCE = `
```
marekkirejczyk: Good idea to simplify tests

sz-piotr:
Q: Shall we use @ethereum-waffle? 
A: Yes

Q: Why not waffle-cli?
A: Makes sense. And later changing it to `@ethereum-waffle/cli` would remove all unnecessary exports

Q: Should waffle-provider be waffle-ganache or waffle-ganache-provider
A: While I think this makes sense no other solution than ganache won't be ready any time soon. And because the packages will be separate we could do a major release only on waffle-provider. So I think sticking with waffle-provider will do two positive things:
  1. Not confuse the users, because currently no other providers exist, also YAGNI
  2. Allow us to maintain our brand, people won't be reminded that it is a thin wrapper around ganache constantly

Q: Need to bump version, before release, I propose 3.0.0-beta
A: Why not 2.3.0? There is not one breaking change here.

Q: Doc + Readme update would be good
A: Yep. Will try to add some info about this.

Q: Why one PR? 😨 Anyway, need a rebase now. 😈
A: After much internal conflict I decided that those changes would be easier to review and merge as a whole even though it is a beast of a PR. Sorry :(

package.json
sz-piotr: Publishing is tricky. There are egde cases (what if one package has no changes? What if a package has changes, but the version is not bumped). I agree that it would be nice, but I see it as a future improvement and not a part of this pr.

waffle-chai/test/contracts/Events.ts
sz-piotr: Yeah, I was sick of everything being compiled every time and needing all of the dependencies

waffle-waffle/LICENSE
sz-piotr: Will change to `Copyright 2020 Ethworks sp z o.o.`

.circleci/config.yml
sz-piotr: Will change, this will also change the name of the the job to `lint` from `lint_and_build` so you will need to change repo configuration to merge it

.gitignore
sz-piotr: It automatically inherits from parent .gitignore. So no need for them.

waffle-compiler/.npmignore
sz-piotr: As per documentation (https://docs.npmjs.com/misc/developers) files starting with `.git` are automatically ignored

waffle-chai/.npmignore
sz-piotr: As per documentation (https://docs.npmjs.com/misc/developers) files starting with `.git` are automatically ignored

waffle-compiler/.npmignore
marekkirejczyk: My understanding is that only ‘.git’ is ignored (not all prefixed)
> By default, the following paths and files are ignored, so there’s no need to add them to .npmignore explicitly:
> (...)
> .git

sz-piotr: Checked this manually :) The gitignore file is not present in the bundle when using `npm pack --dry-run`

pcowgill:
Congrats on merging this!

pcowgill:
Thoughts on putting all of the packages in a `packages/` dir to match `lerna`'s style?

marekkirejczyk:
@pcowgill Somehow we like how you open GitHub and can see the structure instantly :) 
Makes easier to find sth quickly.

PR #184, opened 2020-02-05
Add MockProviderLike

sz-piotr:
This PR will enable the Buidler team to integrate their provider into our legacy API.


sz-piotr:
@alcuadrado

Issue #100, opened 2019-04-11
expect().to.emit.withArgs fails when 2 contract trigger the same event


Amxx:
I have a case where a transaction triggers the same event on 2 contracts with different arg:

→ contractA emits SuccessCall(1)
→ contractB emits SuccessCall(0)

If I try

    expect(...).to.emit(contractA, "SuccessCall").withArgs(1).emit(contractB, "SuccessCall").withArgs(0)

I get an error (from arguments for the call). It looks like the filtering on contract address in not done properly ... I'm also wondering what would happen is a single contract was to emit the same event multiple times with different args.

sz-piotr:
We will be happy to accept a PR that solves this!

rzadp:
- [ ] Same event on 2 contracts
- [ ] Same event on 1 contract with multiple args

PR #189, opened 2020-02-17
Filter logs for the contract in question

rzadp:
Solves one part of https://github.com/EthWorks/Waffle/issues/100


waffle-chai/test/events.ts
```diff
@@ -5,11 +5,14 @@ import {EVENTS_ABI, EVENTS_BYTECODE} from './contracts/Events';
 
 describe('INTEGRATION: Events', () => {
   const [wallet] = new MockProvider().getWallets();
+  let eventFactory: ContractFactory;
   let events: Contract;
 
   beforeEach(async () => {
     const factory = new ContractFactory(EVENTS_ABI, EVENTS_BYTECODE, wallet);
     events = await factory.deploy();
+    eventFactory = new ContractFactory(EVENTS_ABI, EVENTS_BYTECODE, wallet);
```
sz-piotr: There is no need to have two identical factories. Store the one from line 12 in a variable and then use it in the new test

waffle-chai/test/events.ts
```diff
@@ -5,11 +5,14 @@ import {EVENTS_ABI, EVENTS_BYTECODE} from './contracts/Events';
...
+    events = await eventFactory.deploy();
```
sz-piotr: Why deploy this twice?

rzadp: That must have been a mindless copy-paste or unstashing mistake. Please review again.

Issue #187, opened 2020-02-10
revertedWith reports error with actual and expected interchanged


krzysztof-jelski:
With assertion:
`    await expect(contract.callThatShouldBeReverted(().to.be.revertedWith('Insufficient funds');
`
the failing test result is
```
AssertionError: Expected transaction to be reverted
Expected :not reverted
Actual   :reverted
```
Should be the other way around
```
Expected :reverted
Actual   :not reverted
```

PR #191, opened 2020-02-17
Correct mismatched error reports

rzadp:
Closes https://github.com/EthWorks/Waffle/issues/187


rzadp:
> Can you add a comment at the top of the file that explains what the arguments are to this.assert?

Done.

PR #205, opened 2020-03-23
Increase max buffer during native compilation (#179)

bohendo:
Closes #179 (for real this time)
(This problem was also discussed in [solidity issue 8133](https://github.com/ethereum/solidity/issues/8133))

In `waffle-compiler/src/compileNative.ts` we're calling `execSync` and not specifying a `maxBuffer` option so it defaults to `1024 * 1024`. This is a sane default & works for most cases but my org is trying to compile a [bunch of contracts](https://github.com/ConnextProject/indra/tree/staging/modules/contracts/contracts) and either the input or output exceeds this default.

This PR simply specifies a `maxBuffer` value that's 4x this default (we need a value ~2.5x bigger than the default so 4x gives us a fair bit of wiggle room). Because this option specifies a **max** limit on the buffer size, I don't think increasing this limit would have any effect on other projects which are compiling fewer contracts.

PR #212, opened 2020-04-08
Increase max output buffer size during native and docker compilation

msieczko:
Closes #205


marekkirejczyk:
Released in [2.4.1](https://github.com/EthWorks/Waffle/releases/tag/2.4.1).

PR #214, opened 2020-04-10
Bump to 2.4.1

marekkirejczyk:


waffle-provider/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@ethereum-waffle/provider",
   "description": "A mock provider for your blockchain testing needs.",
-  "version": "2.4.0",
+  "version": "2.4.1",
```
sz-piotr, line 4: No changes were made to the provider I think

docs/release-notes/2.4.1.md
```diff
@@ -0,0 +1,3 @@
+## Changes:
```
sz-piotr, line 1: Also mention fixing the bug in big number chai matchers:
previously `to.be.gt` didn't work
also `expect(3).to.be.at.least(utils.bigNumberify(2))` didn't pass

sz-piotr: And this applies to all operations that take the order of arguments into account, so `>` `>=` `<` and `<=`

waffle-provider/package.json
marekkirejczyk: Yeah, it easier for me to just bump all (need to add automatic bumping).

Issue #206, opened 2020-03-25
vyper support


paulperegud:
Does Waffle support contracts written in Vyper? If no, what would it take to add such support?

sz-piotr:
Currently waffle only supports compiling solidity contracts. In order to add vyper support a new sub-package would have to be created, similar to @ethereum-waffle/compiler that deals with vyper and then the cli would have to be able to decide which compiler to use based on the config file

hal909:
+1 For this. It would be amazing to have vyper support for waffle!

marekkirejczyk:
Experimental support for Vyper released in [2.5.0](https://github.com/EthWorks/Waffle/releases/tag/2.5.0).

PR #220, opened 2020-04-15
Add Vyper compilation

rzadp:
- Progresses https://github.com/EthWorks/Waffle/issues/206
- I think next steps could be:
    - Add more `.vy` contracts for more solid tests, including library linking
    - Possibly rewrite to vyper all current `.sol` contracts like `One.sol` or `Two.sol` and then merge `e2e.ts` and `e2e_vyper.ts` together?

### DIFF
--- a/waffle-compiler/src/buildUitls.ts
+++ b/waffle-compiler/src/buildUitls.ts
@@ -8,10 +8,14 @@ export function buildSourcesObject(files: ImportFile[]): Record<string, { conten
   return result;
 }
 
-export function buildInputObject(files: ImportFile[], overrides: any = {}) {
+export function buildInputObject(
+  files: ImportFile[],
+  overrides: any = {},
+  language: 'Solidity' | 'Vyper' = 'Solidity'
+) {
   const sources = buildSourcesObject(files);
   return {
-    language: 'Solidity',
+    language,
     sources,
     settings: {
       outputSelection: {'*': {'*': ['abi', 'evm.bytecode', 'evm.deployedBytecode']}},
--- a/waffle-compiler/src/compileDockerVyper.ts
+++ b/waffle-compiler/src/compileDockerVyper.ts
@@ -1,17 +1,15 @@
-import {join} from 'path';
 import {Config} from './config';
 import {execSync} from 'child_process';
-import {buildInputObject} from './buildUitls';
 import {ImportFile} from '@resolver-engine/imports';
 import {solcOutputMaxBuffer} from './compiler';
+import {buildInputObject} from './buildUitls';
 
-const CONTAINER_PATH = '/home/project';
-const NPM_PATH = '/home/npm';
+const CONTAINER_PATH = '/project';
 
-export function compileDocker(config: Config) {
+export function compileDockerVyper(config: Config) {
   return async function compile(sources: ImportFile[]) {
     const command = createBuildCommand(config);
-    const input = JSON.stringify(buildInputObject(sources, config.compilerOptions), null, 2);
+    const input = JSON.stringify(buildInputObject(sources, config.compilerOptions, 'Vyper'), null, 2);
     return JSON.parse(execSync(command, {input, maxBuffer: solcOutputMaxBuffer}).toString());
   };
 }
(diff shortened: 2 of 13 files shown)


marekkirejczyk:
@sz-piotr Will you find a moment to look at this?

waffle-compiler/src/buildUitls.ts
```diff
@@ -8,10 +8,14 @@ export function buildSourcesObject(files: ImportFile[]): Record<string, { conten
   return result;
 }
 
-export function buildInputObject(files: ImportFile[], overrides: any = {}) {
+export function buildInputObject(
+  files: ImportFile[],
+  overrides: any = {},
+  language: 'Solidity' | 'Vyper' = 'Solidity'
```
marekkirejczyk, line 14: extract type `Language`

waffle-compiler/src/getCompileFunction.ts
```diff
@@ -11,11 +12,13 @@ export type CompileFunction = (
...
 export function getCompileFunction(config: Config): CompileFunction {
   if (config.compilerType === 'native') {
-    return compileNative(config);
+    return compileNativeSolc(config);
   } else if (config.compilerType === 'dockerized-solc') {
-    return compileDocker(config);
+    return compileDockerSolc(config);
   } else if (config.compilerType === 'solcjs') {
     return compileSolcjs(config);
+  } else if (config.compilerType === 'dockerized-vyper') {
```
sz-piotr, line 20: This is super elegant!

waffle-compiler/src/findInputs.ts
```diff
@@ -12,7 +12,7 @@ export function findInputs(sourcePath: string) {
       const filePath = path.join(dir, file);
       if (isDirectory(filePath)) {
         stack.push(filePath);
-      } else if (file.endsWith('.sol')) {
+      } else if (file.endsWith('.sol') || file.endsWith('.vy')) {
```
sz-piotr, line 15: I think I'd prefer having the extension passed as an argument to findInputs. That way we won't have weird errors where we accidentally mix languages. So having `Main.vy` and `Dependency.vy` as well as `Main.sol` and `Dependency.sol` in the same directory wouldn't interfere.

sz-piotr: Also this is untested, because there are no e2e tests where you have vyper code with imports

waffle-compiler/src/compileDockerVyper.ts
```diff
@@ -0,0 +1,27 @@
...
+import {execSync} from 'child_process';
+import {ImportFile} from '@resolver-engine/imports';
+import {solcOutputMaxBuffer} from './compiler';
+import {buildInputObject} from './buildUitls';
+
+const CONTAINER_PATH = '/project';
+
+export function compileDockerVyper(config: Config) {
+  return async function compile(sources: ImportFile[]) {
+    const command = createBuildCommand(config);
```
sz-piotr, line 11: add warning that this is experimental and the api might change without major version change.

rzadp:
> Missing readme and documentation

Readme and docs are missing on purpose - we don't want to announce this feature just yet, keep it experimental to easily do breaking changes if needed.

@sz-piotr Agree?

waffle-compiler/src/config/inputToConfig.ts
```diff
@@ -66,7 +66,7 @@ function validate(config: any): asserts config is Config {
 const checkSourceDirectory = checkType('sourceDirectory', 'string');
 const checkOutputDirectory = checkType('outputDirectory', 'string');
 const checkNodeModulesDirectory = checkType('nodeModulesDirectory', 'string');
-const checkCompilerType = checkEnum('compilerType', ['native', 'dockerized-solc', 'solcjs']);
+const checkCompilerType = checkEnum('compilerType', ['native', 'dockerized-solc', 'solcjs', 'dockerized-vyper']);
```
marekkirejczyk, line 69: Also update in [Config.ts](https://github.com/EthWorks/Waffle/blob/733df25049036482f16f5583183ed97461b58159/waffle-compiler/src/config/config.ts#L15)

marekkirejczyk:
Closing, continue in #222

PR #225, opened 2020-04-20
Display revert location

Jozwiaczek:
image

## TODO

- [x] Try to hook-up error overriding in a simpler way
- [ ] Add a lot more validations so it doesn't crash if something is wrong (eg. bytecode doesn't match)
  - [x] Finds source location for a simple revert in the called method
  - [ ] Finds source location for a revert in a method called from the called method
  - [ ] Finds source location for a revert in an inherited contract
  - [ ] Finds source location for a revert in an inherited contract from node_modules
  - [ ] Finds source location for a revert in a library
  - [x] Does not find a revert location for contracts called by the called contract. (Note that there shouldn't be any error output in the console - this must be a silent fail)
- [x] Have a default build dir (`./build`)
- [ ] Use relative paths in source output (currently absolute - result in nondeterministic builds)
- [ ] Support more RPCs
  * [x] eth_estimateGas
  * [ ] eth_sendTransaction
  * [ ] eth_call
- [ ] (separate PR) Support combined output JSON
- [ ] (optional) Figure out a way to support internal transactions (evm_trace maybe?)


marekkirejczyk:
I propose to first fix linter, than tests, than keep refactoring

sz-piotr:
Let's make sure the CI is green

marekkirejczyk:
Seem to be stalled, closing for now. Feel free to reopen.

PR #230, opened 2020-04-21
Return call arguments history for MockContract

vanruch:


waffle-mock-contract/test/mocha.opts
```diff
@@ -2,3 +2,4 @@
 --timeout 50000
 --no-warnings
 test/**/*.test.{js,ts}
+--exit
```
sz-piotr, line 5: Subscribing to events for mock contracts should not cause leaks

PR #231, opened 2020-04-22
Record call history in MockProvider

sz-piotr:
Closes #230


waffle-provider/src/MockProvider.ts
```diff
@@ -17,4 +41,12 @@ export class MockProvider extends providers.Web3Provider {
   createEmptyWallet() {
     return Wallet.createRandom().connect(this);
   }
+
+  clearCallHistory() {
+    this.callHistory = [];
+  }
+
+  getCallHistory(): readonly RecordedCall[] {
```
vanruch, line 49: replace it with `get callHistory()`?

waffle-provider/test/callHistory.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  it('records failing calls', async () => {
+    const provider = new MockProvider();
+    const [wallet] = provider.getWallets();
+
+    const token = await deployToken(wallet, 10);
+
+    provider.clearCallHistory();
+    try {
+      await token.transfer(wallet.address, 20);
+    } catch {}
```
vanruch, line 89: `await expect(token.transfer(wallet.address, 20)).to.be.reverted`

waffle-provider/src/MockProvider.ts
```diff
@@ -17,4 +41,12 @@ export class MockProvider extends providers.Web3Provider {
...
+    return this.callHistory;
+  }
 }
```
vanruch, line 52: Not sure if this should be put here, but we need some api to filter history by address and decode by ABI

sz-piotr: I would say that this should go in the matchers

waffle-provider/src/MockProvider.ts
```diff
@@ -1,12 +1,36 @@
...
+  readonly address: string;
+  readonly data: string;
+}
+
 export class MockProvider extends providers.Web3Provider {
+  private callHistory: RecordedCall[] = []
+
   constructor(private options?: Ganache.IProviderOptions) {
     super(Ganache.provider({...defaults, ...options}) as any);
+    this.recordCallHistory();
```
marekkirejczyk, line 17: You probably want a extract class `CallHistory`, so that we stick to SRP, and in the constructor you have:
```ts
this.callHistory = new CallHistory();
this.callHistory.record(this._web3Provider);
```

PR #233, opened 2020-04-23
Add Matcher 'to be called'

Jozwiaczek:


waffle-chai/src/matchers/ethCalledWith.ts
```diff
@@ -0,0 +1,32 @@
...
+    const [contract, fnName] = this._obj;
+
+    if (!(contract instanceof Contract)) {
+      throw new TypeError('ethCalled: argument must be a contract');
+    }
+    const provider = contract.provider;
+    if (!(provider instanceof MockProvider)) {
+      throw new TypeError('ethCalled: contract.provider must be a MockProvider');
+    }
+    if (fnName !== undefined && typeof fnName !== 'string') {
```
sz-piotr, line 17: There needs to be another check that this function exists in the contract ABI

waffle-chai/src/matchers/ethCalledWith.ts
```diff
@@ -0,0 +1,32 @@
...
+      throw new TypeError('ethCalled: function name must be a string');
+    }
+
+    const fnSighash = contract.interface.functions[fnName].encode(parameters);
+
+    const isInCallHistory = (cb: IsInCallHistory): boolean => provider.callHistory.some(cb);
```
sz-piotr, line 23: i'd get rid of this function and the `IsInCallHistory` type

waffle-chai/src/matchers/ethCalled.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (!(provider instanceof MockProvider)) {
+      throw new TypeError('ethCalled: contract.provider must be a MockProvider');
+    }
+    if (fnName !== undefined && typeof fnName !== 'string') {
+      throw new TypeError('ethCalled: function name must be a string');
+    }
+
+    const fnSighash = fnName && contract.interface.functions[fnName].sighash;
+
+    const isInCallHistory = (cb: IsInCallHistory): boolean => provider.callHistory.some(cb);
```
sz-piotr, line 30: I'd get rid of this function and the `IsInCallHistory` type

waffle-chai/src/matchers/ethCalled.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    const getAssertionResult = (): AssertionResultInterface => {
```
sz-piotr, line 32: No need to have a function here. You could do an if on `fnSighash` and do different asserts in `if` and `else`. The code will be cleaner and shorter

waffle-chai/src/matchers/ethCalled.ts
```diff
@@ -0,0 +1,56 @@
...
+      if (fnSighash) {
+        return {
+          wasCalled: isInCallHistory(call => call.data === fnSighash),
```
sz-piotr, line 35: this check is flawed. it will work for a function without arguments, but there are two problems.
1) If a function accepts arguments `call.data` contains more stuff, so you need `.startsWith`
2) You also need to check for the address since the function could be called on a different contract

waffle-chai/test/contracts/Events.ts
```diff
@@ -2,59 +2,73 @@ export const EVENTS_SOURCE = `
   pragma solidity ^0.6.0;
 
   contract Events {
```
sz-piotr, line 4: It would probably be better to have a separate contract for testing this new feature specifically

waffle-chai/test/matchers/ethCalledWith.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+async function setup() {
+  const provider = new MockProvider();
+  const [deployer] = provider.getWallets();
+
+  const factory = new ContractFactory(EVENTS_ABI, EVENTS_BYTECODE, deployer);
+  return {contract: await factory.deploy()};
+}
+
+describe('INTEGRATION: ethCalledWith', () => {
```
sz-piotr, line 14: We need more coverage here

waffle-chai/src/matchers/ethCalledWith.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+    const provider = contract.provider;
+    if (!(provider instanceof MockProvider)) {
+      throw new TypeError('ethCalled: contract.provider must be a MockProvider');
+    }
+    if (fnName !== undefined && typeof fnName !== 'string') {
+      throw new TypeError('ethCalled: function name must be a string');
+    }
+
+    const fnSighash = contract.interface.functions[fnName].encode(parameters);
```
sz-piotr, line 21: This is no longer the signature hash. This is now the entire function call data

waffle-chai/src/matchers/ethCalled.ts
```diff
@@ -0,0 +1,56 @@
...
+    const [contract, fnName] = Array.isArray(arg) ? arg : [arg];
+
+    if (!(contract instanceof Contract)) {
+      throw new TypeError('ethCalled: argument must be a contract');
+    }
+    const provider = contract.provider;
+    if (!(provider instanceof MockProvider)) {
+      throw new TypeError('ethCalled: contract.provider must be a MockProvider');
+    }
+    if (fnName !== undefined && typeof fnName !== 'string') {
```
sz-piotr, line 24: There needs to be another check that this function exists in the contract ABI.

waffle-chai/src/index.ts
```diff
@@ -19,4 +21,6 @@ export function waffleChai(chai: Chai.ChaiStatic, utils: Chai.ChaiUtils) {
   supportProperHex(chai.Assertion);
   supportChangeBalance(chai.Assertion);
   supportChangeBalances(chai.Assertion);
+  supportEthCalled(chai.Assertion);
+  supportEthCalledWith(chai.Assertion);
```
sz-piotr, line 25: I would consider having those two in the same file. We could then extract the common parts into a function, for example for data type checking

waffle-chai/src/matchers/ethCalled/calledOnContractValidators.ts
```diff
@@ -0,0 +1,39 @@
+import {Contract} from 'ethers';
+import {MockProvider} from '@ethereum-waffle/provider';
+
+export function validateExpectArguments(contract: any) {
```
sz-piotr, line 4: Rename to validateContract

waffle-chai/src/matchers/ethCalled/calledOnContract.ts
```diff
@@ -0,0 +1,28 @@
...
+    validateExpectArguments(contract);
+    if (fnName !== undefined) {
+      validateFnName(fnName, contract);
+    }
+
+    const fnSighash = contract.interface.functions[fnName].sighash;
+    const callHistory = contract.provider.callHistory;
+
+    this.assert(
+      callHistory.some((call: RecordedCall) => {
```
sz-piotr, line 17: why no one line and without the explicit type and an additional variable
```ts 
contract.provider.callHistory.some(
  call => call.address === contract.address && call.data.startsWith(fnSighash)
)
```

waffle-chai/src/matchers/ethCalled/calledOnContractValidators.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function isInstanceOfContract(val: any): asserts val is Contract {
+  if (!(val instanceof Contract)) {
+    throw new TypeError('ethCalled: argument must be a contract');
+  }
+}
+
+function isInstanceOfMockProvider(val: any): asserts val is MockProvider {
+  if (!(val instanceof MockProvider)) {
```
sz-piotr, line 30: use validateCondition

waffle-chai/src/matchers/ethCalled/calledOnContractValidators.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function isString(val: any): asserts val is string {
+  if (typeof val !== 'string') {
+    throw new TypeError('ethCalled: function name must be a string');
+  }
+}
+
+function isInstanceOfContract(val: any): asserts val is Contract {
+  if (!(val instanceof Contract)) {
```
sz-piotr, line 24: use validateCondition

waffle-chai/src/matchers/ethCalled/calledOnContractValidators.ts
```diff
@@ -0,0 +1,39 @@
...
+export function validateFnName(fnName: string, contract: any) {
+  isString(fnName);
+  validateCondition(
+    fnName in contract.interface.functions,
+    'ethCalled: function must exist in provided contract'
+  );
+}
+
+function isString(val: any): asserts val is string {
+  if (typeof val !== 'string') {
```
sz-piotr, line 18: use validateCondition

waffle-chai/src/matchers/ethCalled/calledOnContractWith.ts
```diff
@@ -0,0 +1,25 @@
...
+  Assertion.addMethod('calledOnContractWith', function (this: any, contract: any, parameters: Array<any>) {
+    const fnName = this._obj;
+
+    validateExpectArguments(contract);
+    validateFnName(fnName, contract);
+
+    const funCallData = contract.interface.functions[fnName].encode(parameters);
+
+    this.assert(
+      contract.provider.callHistory.some((call: RecordedCall) => {
```
sz-piotr, line 14: same, consider a one line solution

waffle-chai/src/types.ts
```diff
@@ -12,6 +12,8 @@ declare namespace Chai {
     properAddress: void;
     changeBalance(wallet: any, balance: any): AsyncAssertion;
     changeBalances(wallets: any[], balances: any[]): AsyncAssertion;
+    calledOnContract(contract: any): void;
+    calledOnContractWith(contract: any, ...parameters: any[]): void;
```
sz-piotr, line 16: This type is incorrect or the code is incorrect. The tests use it as `calledOnContractWith(contract, [1, 2]` so the type should be `calledOnContractWith(contract: any, parameters: any[])`

waffle-chai/src/matchers/ethCalled/calledOnContractWith.ts
```diff
@@ -0,0 +1,25 @@
+import {validateExpectArguments, validateFnName} from './calledOnContractValidators';
+import {RecordedCall} from '@ethereum-waffle/provider';
+
+export function supportCalledOnContractWith(Assertion: Chai.AssertionStatic) {
+  Assertion.addMethod('calledOnContractWith', function (this: any, contract: any, parameters: Array<any>) {
```
sz-piotr, line 5: use `any[]` for consistency

sz-piotr: remove `this: any`

waffle-chai/src/matchers/ethCalled/calledOnContract.ts
```diff
@@ -0,0 +1,28 @@
+import {validateExpectArguments, validateFnName} from './calledOnContractValidators';
+import {RecordedCall} from '@ethereum-waffle/provider';
+
+export function supportCalledOnContract(Assertion: Chai.AssertionStatic) {
+  Assertion.addMethod('calledOnContract', function (this: any, contract: any) {
```
sz-piotr, line 5: remove `this: any`

waffle-chai/test/matchers/ethCalled/calledOnContractValidators.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  it('throws type error when the provided function is not in the contract', async () => {
+    const {contract} = await setup();
+
+    expect(
+      () => expect('notExistingFunction').to.be.calledOnContract(contract)
+    ).to.throw(TypeError, 'ethCalled: function must exist in provided contract');
+  });
+
+  it('throws type error when the provided function is from another contract', async () => {
```
sz-piotr, line 50: This test is no different than the previous. There should be a test where you deploy two identical contracts, call the function on one and test that it was called on the one and wasn't called on the other

waffle-chai/test/matchers/ethCalled/calledOnContractWith.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  it('throws assertion error when contract function with parameter was called', async () => {
+    const {contract} = await setup();
+    await contract.callWithParameter(2);
+
+    expect(
+      () => expect('callWithParameter').not.to.be.calledOnContractWith(contract, [2])
+    ).to.throw(AssertionError, 'Expected contract function with parameters NOT to be called');
+  });
+});
```
sz-piotr, line 63: also we need a test where it is called twice with different args and it still finds the one call we wanted

waffle-chai/src/matchers/ethCalled/calledOnContract.ts
Jozwiaczek: @sz-piotr, I added type because without it there is TS warning. Should I leave RecordedCall type?

sz-piotr: Oh it is actually a deeper problem. Typescript does not understand `validateExpectArguments` and thinks contract is `any`. You should validate the contract directly:
```
validateContract(contract);
validateMockProvider(contract.provider);
```

sz-piotr: Then you can remove the `RecordedCall` type from the callback

sz-piotr:
Also rename the folder `ethCalled` to `calledOnContract`

PR #234, opened 2020-04-29
Cleanup compiler code

sz-piotr:


waffle-compiler/test/compiler/e2e.ts
```diff
@@ -131,7 +131,9 @@ describe('E2E: Compiler integration', async () => {
         const [wallet] = new MockProvider().getWallets();
         const libraryPath = resolve(join(configuration.outputDirectory, 'MyLibrary.json'));
         const MyLibrary = require(libraryPath);
-        const LibraryConsumer = deepCopy(require(resolve(join(configuration.outputDirectory, 'Two.json'))));
```
marekkirejczyk, line 134: I would argue `deepCopy(...)` is better than `JSON.parse(JSON.stringify(...))`.
Shorter, states the intention. We can take it from loadash (however why add dependency?)

PR #236, opened 2020-04-29
Fix imports from undeclared modules

sz-piotr:


.eslintrc.json
```diff
@@ -330,9 +342,15 @@
...
-    "no-restricted-imports": ["error", {
-      "patterns": ["original-works-core/src", "original-works-core/dist"]
-    }],
+    "no-restricted-imports": [
+      "error",
+      {
+        "patterns": [
+          "original-works-core/src",
+          "original-works-core/dist"
+        ]
```
marekkirejczyk, line 351: eee... are you sure it is from this project?

sz-piotr: hahaha, if you look closely All I did here is reformat, so that was here before. Not sure how it got here.

PR #237, opened 2020-04-30
Migrate to ethers v5

### LINKED ISSUE #162: ethers v5 support
I see in [the docs](https://ethereum-waffle.readthedocs.io/en/latest/) that ethers v4 is supported in `ethereum-waffle` 2.x.

With [ethers v5 ready to use for new projects](https://github.com/ethers-io/ethers.js/pull/621), I thought it would be good to kick off this conversation here.

pedrouid:
fix #162


pedrouid:
cc @marekkirejczyk @sz-piotr

marekkirejczyk:
I was able to diagnose two problems (probably there is more):

### Problem 1
 `parseLog` method it not part of `Contract.interface` anymore (didn't find where is it now, thou)


### Problem 2
Contract.interface.events have different shape now. It was indexed with just the name of the event (e `Two`), not it is fully qualified ABI name (e.g. `Two(uint256,string)`).

I wrote a quick function to replace [in emit.ts line 13](https://github.com/EthWorks/Waffle/blob/60c67245730a1ed110323936a662137fb45068b2/waffle-chai/src/matchers/emit.ts#L13):
```ts
const eventDescription = contract.interface.events[eventName];
```
With:
```ts
const eventDescription = findEventFragment(contract, eventName);
```

Links to `findEventFragment` [function](https://gist.github.com/marekkirejczyk/3f03b10117469307e71a50f6d791ccb3) and [tests](https://gist.github.com/marekkirejczyk/fcc6d98c8b012863351384dc17fd56d1).

(Would be good to work on types)

### Next steps
Would be wonderful if you want to help with those two problems. We will be focusing on releasing 2.5 early next week, we can help more in the second half of the week.

pedrouid:
Hey @marekkirejczyk, thanks for reviewing.

Regarding Problem 1, parseLog is still available but a different API, I will update the PR for this.

Regarding Problem 2, good catch that was my bad because the Interface API already includes the getEvent method which I should've used to get EventFragment, I will push an update for this too.

PS - eventDescription has been renamed to eventFragment on v5

pedrouid:
Also fixed how `validateFnname` asserts functions exist in the Contract interface

We are now done to 7 failing tests 🙌 The remaining are all due to mismatching error messages

pedrouid:
I was able to get it down to 2 failing tests but on the `supportReverted` and `supportedRevertedWith` I had to catch ethers.js error messages which were being thrown for `UNPREDICTABLE_GAS_LIMIT`

pedrouid:
Finally I ignore the `getEvent` error message and instead evaluated the eventFragment to be undefined so that that the `emit` assertion could be resolved correctly

pedrouid:
All test are passing! cc @marekkirejczyk @sz-piotr 

I would love to get a pre-release for this migration if possible 🙏

pedrouid:
> I was able to get it down to 2 failing tests but on the `supportReverted` and `supportedRevertedWith` I had to catch ethers.js error messages which were being thrown for `UNPREDICTABLE_GAS_LIMIT`

I've reached out to @ricmoo about this issue and he shared that this issue would be present in v4 as well. This error is triggered when the gasLimit is not provided and the estimateGas fails because either the transaction reverts or the node failed to estimate gas

Since I'm sure if this issue is local to my machine, I will revert that change for `supportReverted` and `supportedRevertedWith` to test it in CI

pedrouid:
CI had the same behavior, these client errors could be asserted targetting each code to ensure only to catch transaction reverts but for this PR I'm catching any client error assuming the user will only use `reverted` and `reveredWith` for contract calls

marekkirejczyk:
@pedrouid Great work, I allocated time tomorrow around noon for releasing. Will review and if there are not show stoppers, I'll release beta version

waffle-chai/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@ethereum-waffle/provider": "^2.4.1",
-    "ethers": "^4.0.45"
+    "ethers": "^5.0.0-beta.184"
```
marekkirejczyk, line 39: Why `.184` not `.1`?

waffle-chai/src/matchers/emit.ts
```diff
@@ -1,18 +1,23 @@
-import {Contract} from 'ethers';
+import {Contract, providers, utils} from 'ethers';
 
 export function supportEmit(Assertion: Chai.AssertionStatic) {
-  const filterLogsWithTopics = (logs: any[], topic: any, contractAddress: string) =>
+  const filterLogsWithTopics = (logs: providers.Log[], topic: any, contractAddress: string) =>
```
marekkirejczyk, line 4: 👍

waffle-chai/src/matchers/emit.ts
```diff
@@ -1,18 +1,23 @@
...
     const derivedPromise = promise.then((tx: any) =>
       contract.provider.getTransactionReceipt(tx.hash)
-    ).then((receipt: any) => {
-      const eventDescription = contract.interface.events[eventName];
+    ).then((receipt: providers.TransactionReceipt) => {
+      let eventFragment: utils.EventFragment | undefined;
+      try {
+        eventFragment = contract.interface.getEvent(eventName);
+      } catch (e) {
+        // ignore error
```
marekkirejczyk, line 17: Good rule of the tumb is to rather throw an exception than let it silently fail.

marekkirejczyk:
I have read 40%+, will continue in the evening. So far looks very good! 👍 
@pedrouid Would you be able to do a rebase?

waffle-chai/src/matchers/emit.ts
pedrouid: 👌

pedrouid:
Just rebased! Looking forward for the rest of the review... very excited to be moving this to beta so that I can unblock Connext's monorepo migration

marekkirejczyk:
@pedrouid 
Very excited too. Aim to release today. I am reviewing now. Test are failing after migration. Would be good to fix.

pedrouid:
It was an easy fix, since waffle-ens was using ethers v4

marekkirejczyk:
Hopefully last think @Pedro:
I just released 2.5.0, which will now go on the branch and I can now merge it to master, we will need to change package.json versions to 3.0.0-beta.1 (need a rebase)

waffle-chai/src/matchers/reverted.ts
```diff
@@ -1,31 +1,31 @@
...
+      this.assert(
+        isReverted || isThrown || isError,
+        `Expected transaction to be reverted, but other exception was thrown: ${error}`,
+        'Expected transaction NOT to be reverted',
+        'Transaction reverted.',
+        error
+      );
+      return error;
+    };
+    const derivedPromise = promise.then(onSuccess, onError);
```
marekkirejczyk, line 28: Good long needed refactoring 👍

waffle-mock-contract/test/direct.test.ts
```diff
@@ -13,27 +13,27 @@ describe('Mock Contract - Integration (called directly)', () => {
   const [wallet] = new MockProvider().getWallets();
 
   it('throws readable error if mock was not set up for a method', async () => {
-    const mockCounter = await deployMockContract(wallet, Counter.interface);
+    const mockCounter = await deployMockContract(wallet, Counter.abi);
```
marekkirejczyk, line 16: @sz-piotr Hmm, is that right?

sz-piotr: Yeah, interface is the legacy way of accessing it. Abi has many advantages. Shorter, standard and also isn't a reserved word

marekkirejczyk: thx

waffle-chai/test/matchers/events.test.ts
```diff
@@ -30,134 +33,199 @@ describe('INTEGRATION: Events', () => {
...
+        'One',
+        '0x0000000000000000000000000000000000000000000000000000000000000001'
+      )
       .and.to.emit(events, 'Two');
   });
 
   it('Event with proper args', async () => {
-    await (expect(events.emitOne()).to.emit(events, 'One'))
-      .withArgs(1, 'One', '0x00cfbbaf7ddb3a1476767101c12a0162e241fbad2a0162e2410cfbbaf7162123');
+    await expect(events.emitOne())
```
marekkirejczyk, line 99: Good catch.

waffle-chai/test/matchers/events.test.ts
```diff
@@ -30,134 +33,199 @@ describe('INTEGRATION: Events', () => {
...
+            '0x00cfbbaf7ddb3a1476767101c12a0162e241fbad2a0162e2410cfbbaf7162123',
+            '0x00cfbbaf7ddb3a1476767101c12a0162e241fbad2a0162e2410cfbbaf7162124'
+          ]
+        )
     ).to.be.eventually.rejectedWith(
       AssertionError,
       // eslint-disable-next-line no-useless-escape
-      'Expected \"0\" ' +
-      'to be equal 1'
+      'Expected "0" ' + 'to be equal 1'
```
marekkirejczyk, line 228: Could be merged into one string.

pedrouid:
Alright going to resolve the waffle-ens failing test and merge in those requested changes...

pedrouid:
It looks like Contract is being instantiated correctly but then the method `setAddr` is not available in the interface, I found this error message in the test runtime:

```
Duplicate definition of addr (addr(bytes32), addr(bytes32,uint256))
Duplicate definition of setAddr (setAddr(bytes32,uint256,bytes), setAddr(bytes32,address))
```

Didn't you see this before with ethers v4? @marekkirejczyk

ricmoo:
If the ABI contains conflicting signatures, it is not available by name, only by signature. In v4, the first was chosen, which can lead to incorrect execution.

My guess is the ABI looks like `[ “setAddr(bytes32, address)”, “setAddr(bytes32, uint coinType, bytes)” ]`?

So, calling `contract.setAddr(...)` is ambiguous and it isn’t safe to pick either one. Instead you need to use the signature (e.g. `contract["setAddr(bytes32,address)"]`.

(People next suggest base it off the argument count; this also isn’t safe or possible in JavaScript, in general, and it would confusing why really specific cases work, but others do not... I’ll make a note of that in the docs. ;))

I have a change that will not log the warning, but fighting with the CI right now. Hoping to get that out today though. :)

Make sense?

pedrouid:
Amazing! Thanks for the responsiveness as always @ricmoo 🙌  I've fixed the test!

Could we get this merged? @marekkirejczyk

marekkirejczyk:
Awesome! Will merge it and publish later today.

PR #252, opened 2020-05-08
Add waffle-e2e

Jozwiaczek:


waffle-e2e/buildTestContracts.ts
```diff
@@ -0,0 +1,15 @@
...
+  const compilerVersion = '0.5.15';
+
+  await compileAndSave({
+    sourceDirectory,
+    outputDirectory,
+    compilerVersion
+  });
+};
+
+buildExampleContracts();
```
sz-piotr, line 15: Actually you can just remove this function and just have this called directly
```ts
compileAndSave({
  sourceDirectory: './src',
  outputDirectory: './build',
  compilerVersion: '0.5.15'
})
```

waffle-e2e/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "MIT",
+  "scripts": {
+    "build": "ts-node buildTestContracts",
+    "test:nobuild": "export NODE_ENV=test && mocha",
+    "test": "yarn build && yarn test:nobuild",
+    "lint": "eslint '{src,test}/**/*.ts'",
+    "lint:fix": "eslint --fix '{src,test}/**/*.ts'"
+  },
+  "dependencies": {
+    "@ethereum-waffle/compiler": "^3.0.0-beta.1"
```
sz-piotr, line 14: This project should probably only have dev dependencies

Jozwiaczek: I left it here because of linter warning in buildTestContracts

sz-piotr: add a linter comment in the script then to disable the error

sz-piotr: The linter is configured so that only tests can use dependencies from devDependencies

PR #260, opened 2020-05-13
Set reverse address

VladStarostenko:


waffle-ens/src/EnsBuilder.ts
```diff
@@ -110,4 +110,8 @@ export class ENSBuilder {
     await this.ensureDomainExist(decodedRootNode, options);
     await this.setAddressNonRecursive(domain, address);
   }
+
+  async setReverseAddress(address: string) {
```
marekkirejczyk, line 114: address -> name?

waffle-ens/src/EnsBuilder.ts
```diff
@@ -110,4 +110,8 @@ export class ENSBuilder {
     await this.ensureDomainExist(decodedRootNode, options);
     await this.setAddressNonRecursive(domain, address);
   }
+
+  async setReverseAddress(name: string) {
```
marekkirejczyk, line 114: setReverseAddress -> setName

PR #269, opened 2020-05-18
Fix problem with flat contract import order

vanruch:
Replaces #268

### DIFF
--- a/waffle-compiler/src/flattener.ts
+++ b/waffle-compiler/src/flattener.ts
@@ -54,15 +54,18 @@ function saveToFile(
   fileSystem.createDirectory(outputDirectory);
 
   output.map((contract: Array<GatheredContractInterface>) => {
-    const fileName = path.parse(contract[contract.length - 1].url).base;
+    const rootContract = contract[contract.length - 1];
+    const fileName = path.parse(rootContract.url).base;
     const filePath = path.join(outputDirectory, fileName);
-    const source = ''.concat(...unique(contract.map(replaceDirectivesWithComments(contract[contract.length - 1]))));
+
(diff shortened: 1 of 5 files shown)


waffle-compiler/src/flattener.ts
```diff
@@ -52,31 +54,30 @@ function saveToFile(
...
     fileSystem.writeFile(filePath, source);
   });
 }
 
-function replaceDirectivesWithComments(
-  dependency: GatheredContractInterface,
-  contract: Array<GatheredContractInterface>,
-  source: string
-) {
+function replaceDirectivesWithComments(contract: GatheredContractInterface) {
```
marekkirejczyk, line 65: `contract` -> `rootContract`

waffle-compiler/src/flattener.ts
```diff
@@ -52,31 +54,30 @@ function saveToFile(
   fileSystem.createDirectory(outputDirectory);
 
   output.map((contract: Array<GatheredContractInterface>) => {
-    const fileName = path.parse(contract[0].url).base;
+    const fileName = path.parse(contract[contract.length - 1].url).base;
     const filePath = path.join(outputDirectory, fileName);
-    let source = '';
+    const source = ''.concat(...unique(contract.map(replaceDirectivesWithComments(contract[contract.length - 1]))));
```
marekkirejczyk, line 59: Can we extract a bit?
```ts
const rootContract = contract[contract.length - 1];
const contractsWithDirectivesCommented = contract.map(replaceDirectivesWithComments(rootContract));
const source = ''.concat(...unique(contractsWithDirectivesCommented));
```

PR #280, opened 2020-05-25
Ens example

VladStarostenko:


examples/ens/package.json
```diff
@@ -0,0 +1,48 @@
+{
```
marekkirejczyk, line 1: Tool long, you can use `example-mock-contract` as a start, note: no need to have `build` at all

examples/ens/test/ENS.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  const provider = new MockProvider();
+  const [wallet] = provider.getWallets();
+  let ens: ENS;
+
+  beforeEach(async () => {
+    ens = await deployENS(wallet);
+    await ens.createTopLevelDomain('test');
+  });
+
+  describe('Create domain', async () => {
```
marekkirejczyk, line 21: Let's use more e2e examples, ie:
- createDomain
- setAddress
- setAddressWithReverse

PR #288, opened 2020-05-27
Called-on-contract example

VladStarostenko:


marekkirejczyk:
Should be a bit different:
- two contracts AmIRichAlready and ERC20
- expect('balanceOf').to.be.calledOn(token)

examples/called-on-contract/src/AmIRichAlready.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity ^0.6.2;
+
+interface ERC20 {
+    function balanceOf(address account) external view returns (uint256);
+}
```
marekkirejczyk, line 5: Interface not needed, you can use `ERC20` from `openzeppelin`

Issue #289, opened 2020-05-27
Add function signature lookup to Mock Contracts


asselstine:
The new mock contracts feature is excellent.  However, I've encountered a roadblock:

The mock functions are keyed only on the function name, so if there are multiple functions with the same name only one is available.

This is ideal in most situations, but it would be great if we could also key on the full function signature (like ethers).   For example:

```
await token.mock.mint.withArgs(user1, 10000)
// can be 
await token.mock['mint(address,uint256)'].withArgs(user1, 10000)
```

This would be a very small and beneficial change.

marekkirejczyk:
Released in Beta 3.0.0-beta.3, looking forward to your feedback.

PR #290, opened 2020-05-27
Added ability to lookup mock function by signature.

asselstine:
Addresses issue https://github.com/EthWorks/Waffle/issues/289.

Copied here:

The mock functions are keyed only on the function name, so if there are multiple functions with the same name only one is available.

This is ideal in most situations, but it would be great if we could also key on the full function signature (like ethers). For example:

```javascript
await token.mock.mint.withArgs(user1, 10000)
// can be 
await token.mock['mint(address,uint256)'].withArgs(user1, 10000)
```

This would be a very small and beneficial change.

### DIFF
--- a/waffle-mock-contract/src/index.ts
+++ b/waffle-mock-contract/src/index.ts
@@ -30,11 +30,14 @@ function createMock(abi: ABI, mockContractInstance: Contract) {
   const {functions} = new utils.Interface(abi);
   const encoder = new utils.AbiCoder();
 
-  return Object.values(functions).reduce((acc, func) => ({
-    ...acc,
-    [func.name]: stub(mockContractInstance, encoder, func),
-    [func.format()]: stub(mockContractInstance, encoder, func)
-  }), {} as MockContract['mock']);
+  return Object.values(functions).reduce((acc, func) => {
+    let stubbed = stub(mockContractInstance, encoder, func);
+    return {
+      ...acc,
+      [func.name]: stubbed,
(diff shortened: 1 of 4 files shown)


waffle-mock-contract/src/index.ts
```diff
@@ -32,7 +32,8 @@ function createMock(abi: ABI, mockContractInstance: Contract) {
 
   return Object.values(functions).reduce((acc, func) => ({
     ...acc,
-    [func.name]: stub(mockContractInstance, encoder, func)
+    [func.name]: stub(mockContractInstance, encoder, func),
+    [func.format()]: stub(mockContractInstance, encoder, func)
```
sz-piotr, line 36: the stub could only be created once and assigned twice, i.e. reuse `stub(mockContractInstance, encoder, func)`

waffle-mock-contract/test/contract.test.ts
```diff
@@ -48,6 +48,17 @@ describe('Doppelganger - Contract', () => {
       expect(await pretender.add(5)).to.equal(returnedValue);
     });
 
+    it('allows function to be looked up by signature', async () => {
```
sz-piotr, line 51: A test for overloaded functions would also be good

asselstine:
@sz-piotr updated:

- Added to readme
- Shared 'stubbed' reference for both signature and name
- Improved test

sz-piotr:
Thank you for this PR, very helpful change overall <3

PR #291, opened 2020-05-28
Remove legacy config

sz-piotr:


waffle-compiler/test/config.ts
```diff
@@ -2,143 +2,23 @@ import {expect} from 'chai';
...
+    const input = {
+      name: 'name',
+      compilerType: 'native',
+      compilerOptions: {},
+      foo: 'bar',
+      boo: 123
+    };
+    const output = {
+      compilerType: 'native',
+      compilerOptions: {}
```
marekkirejczyk, line 15: non-empty `compilerOptions` would be more expressive example :P

PR #293, opened 2020-05-29
Add change balance example

VladStarostenko:


examples/change-balance/test/ChangeBalance.test.ts
```diff
@@ -0,0 +1,28 @@
...
+        value: 200
+      })
+    ).to.changeBalance(walletTo, 200);
+  });
+
+  it('Change balance of receiver and sander wallets', async () => {
+    await expect(() =>
+      walletFrom.sendTransaction({
+        to: walletTo.address,
+        gasPrice: 0,
```
sz-piotr, line 23: not zero and take it into account

PR #296, opened 2020-06-01
Add minimal output option

VladStarostenko:


waffle-compiler/test/compiler/outputType.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  it('Create minimal output file with humanReadableAbi', async () => {
+    await compileProject('config-minimal-humanReadableAbi.json');
+    const filePath = join('./build', 'Constantinople.json');
+    const actualOutput = JSON.parse(readFileContent(filePath));
+
+    expect(actualOutput).to.deep.equal({
+      ...expectedMinimalOutput,
+      humanReadableAbi: [
```
sz-piotr, line 54: So my idea was that since it's minimal it looks like this:
```ts
// config
{ outputType: 'minimal' }
// output
{ abi: {...}[], bytecode: string }
```
```ts
// config
{ outputType: 'minimal', outputHumanReadableAbi: true }
// output
{ abi: string[], bytecode: string }
```
instead of
```ts
// config
{ outputType: 'minimal', outputHumanReadableAbi: true }
// output
{ abi: {...}[], humanReadableAbi: string[], bytecode: string }
```

sz-piotr:
Also there should be a mention about this new setting in the docs

waffle-compiler/src/saveOutput.ts
```diff
@@ -49,6 +49,28 @@ export async function saveOutput(
...
+async function saveOutputMinimal(output: any, config: Config, filesystem = fsOps) {
+  for (const [, file] of Object.entries<any>(output.contracts)) {
+    for (const [contractName, contractJson] of Object.entries<any>(file)) {
+      const filePath = join(config.outputDirectory, `${contractName}.json`);
+      filesystem.writeFile(filePath, getMinimalContent(contractJson, config));
+    }
+  }
+}
+
+function getMinimalContent(contractJson: ContractJson, config: Config) {
```
sz-piotr, line 67: This implementation is not future proof. If solc ever outputs more stuff then you are left with more.
How about this:
```ts
const abi = config.outputHumanReadableAbi
  ? getHumanReadableAbi(contractJson.abi)
  : contractJson.abi;
const bytecode = contractJson.evm.bytecode.object;
return JSON.stringify({ abi, bytecode }, null, 2);
```

PR #299, opened 2020-06-02
Add docs job to CI

VladStarostenko:


.circleci/config.yml
```diff
@@ -133,6 +133,23 @@ jobs:
       - run: yarn lint
       - run: yarn test
 
+  docs:
+    docker:
+      - image: circleci/python:latest
+    steps:
+      - checkout
+      - setup_remote_docker
```
sz-piotr, line 141: This shouldn't be needed. We need this for waffle tests, because waffle runs solc on docker

.circleci/config.yml
```diff
@@ -133,6 +133,23 @@ jobs:
...
+      - run:
+          name: Install dependencies
+          command: |
+            cd docs
```
sz-piotr, line 145: you should set the working directory at the top of the job config, like in `example_ens`

marekkirejczyk: not yet done

PR #302, opened 2020-06-03
Add recognising waffle.json without arguments

VladStarostenko:


waffle-compiler/src/config/loadConfig.ts
```diff
@@ -2,6 +2,7 @@ import path from 'path';
 import {InputConfig} from './config';
 
 export async function loadConfig(configPath?: string): Promise<InputConfig> {
+  configPath = configPath || 'waffle.json';
   if (configPath) {
```
marekkirejczyk, line 6: always true

waffle-compiler/src/config/loadConfig.ts
```diff
@@ -2,6 +2,7 @@ import path from 'path';
 import {InputConfig} from './config';
 
 export async function loadConfig(configPath?: string): Promise<InputConfig> {
+  configPath = configPath || 'waffle.json';
```
marekkirejczyk, line 5: Only assign configPath if `waffle.json` exists

marekkirejczyk: A test would be nice

waffle-compiler/test/projects/defaultConfigFile/contracts/Constantinople.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.5.1;
+
+
+contract Constantinople {
```
marekkirejczyk, line 4: Let's use some other example with solidity 0.6.* and nothing to do with constantinopole

PR #306, opened 2020-06-05
docs: update matchers.rst with configuration info

PaulRBerg:


docs/source/matchers.rst
```diff
@@ -1,7 +1,16 @@
 Chai matchers
 =============
 
-A set of sweet chai matchers, makes your test easy to write and read. Below is the list of available matchers:
+A set of sweet chai matchers, makes your test easy to write and read. Before you can start using the matchers, you have to tell chai to use the solidity plugin:
```
marekkirejczyk, line 4: Good catch. I think this only applies to bignumber matchers thou, so we should move it to that section

PaulRBerg: Yeah, just tested it now. I could either edit this PR or let you close it and mention it wherever you think makes sense most.

marekkirejczyk: @sz-piotr Made me realize this is correct.

PR #307, opened 2020-06-05
Add migration guides

VladStarostenko:


docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
+Migration guides
```
marekkirejczyk, line 1: Let's do it a bit different:
- no tabs this time

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
...
+================
+
+Migration from Waffle 2.*.* to Waffle 2.3.0 and Waffle 2.4.0
+------------------------------------------------------------
```
marekkirejczyk, line 5: Migrate from 2.4 to 2.5

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
...
+        sourceDirectory: "...",
+        outputDirectory: "...",
+        nodeModulesDirectory: "...",
+        compilerType: "...",
+        compilerVersion: "...",
+        compilerAllowedPaths: "...",
+        compilerAllowedPaths: "..."
+      }
+
+Migration from Waffle v2 to Waffle v3
```
marekkirejczyk, line 85: From 2.5.* to 3.0.0

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
...
+In :code:`Waffle 3.0.0` we swapped arguments for Fixture, because the provider argument is very rarely used compared to wallets.
+So such implantation should be more convenient for users.
+
+.. tabs::
+
+  .. group-tab:: Waffle 2.5.0
+
+    .. code-block:: ts
+
+      type Fixture<T> = (provider: MockProvider, wallets: Wallet[]) => Promise<T>;
```
marekkirejczyk, line 105: Let's bring here example fixture here instead.
Let's  add example for fixtures

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
...
+    .. code-block:: text
+
+      {
+        sourceDirectory: "...",
+        outputDirectory: "...",
+        nodeModulesDirectory: "...",
+        compilerType: "...",
+        compilerVersion: "...",
+        compilerAllowedPaths: "...",
+        compilerAllowedPaths: "..."
```
marekkirejczyk, line 82: duplicate (last line)

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
...
+We added MockProviderOptions. It will be convenient in the future, when the provider may need some
+other options than :code:`ganacheOptions`.
+
+.. tabs::
+
+  .. group-tab:: Waffle 2.5.0
+
+    .. code-block:: ts
+
+      constructor(private options?: MockProviderOptions) {
```
marekkirejczyk, line 161: examples here needed instead of sourc code

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,217 @@
...
+We dropped support for contract.interface, because it duplicated contract ABI. Now we support just :code:`contract.abi`.
+
+.. tabs::
+
+  .. group-tab:: Waffle 2.5.0
+
+    .. code-block:: text
+
+      {
+        "abi": [
```
marekkirejczyk, line 193: no abi here (interface is a equivalent)

docs/source/migration-guides.rst
marekkirejczyk: What about `ganacheOptions`?

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
+Migration guides
+================
+
+Migration from Waffle 2.2.0 to Waffle 2.3.0
+-------------------------------------------
+
+**Created monorepo**
+
+Packages :code:`waffle-cli`, :code:`waffle-chai`, :code:`waffle-provider`, :code:`waffle-compiler` were migrated to monorepo.
```
sz-piotr, line 9: This is not helpful for outside users. We should say this differently:
```
Waffle was internally migrated to a monorepo. Because of this you can now use parts of waffle individually. We provide the following packages:
- :code:`ethereum-waffle` - core package exporting everything
- :code:`ethereum-compiler` - compile your contracts programmatically
- :code:`ethereum-chai` - chai matchers for better unit testing
- :code:`ethereum-provider` - mock provider to interact with an in-memory blockchain
```

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+
+**Created MockProvider class**
+
+We added MockProvider class, it was change the way to create the provider.
```
sz-piotr, line 13: Improve english

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+  await createMockProvider(options);
+
+*Waffle 2.3.0*
+
+.. code-block:: ts
+
+  provider = new MockProvider();
+
+
+:code:`new MockProvider` returned :code:`providers.Web3Provider`, which is the parent class of :code:`MockProvider`.
```
sz-piotr, line 28: How is this useful for the reader? What do they have to do? What are the consequences?

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
+Migration guides
+================
+
+Migration from Waffle 2.2.0 to Waffle 2.3.0
+-------------------------------------------
+
+**Created monorepo**
```
sz-piotr, line 7: Why use `**` instead of headings like everywhere else? For level 3 you can use `~` https://docutils.sourceforge.io/docs/user/rst/quickstart.html#sections

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+.. code-block:: ts
+
+  wallets = new MockProvider().getWallets()
+
+Migration from Waffle 2.3.0 to Waffle 2.4.0
+-------------------------------------------
+
+**Renamed configuration options**
+
+We renamed configuration options to compile contracts.
```
sz-piotr, line 49: This whole section is very unhelpful. It should go over each of the old keys and state what needs to be changed.

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+Migration from Waffle 2.5.* to Waffle 3.0.0
+-------------------------------------------
+
+There are some new functionality and some slight refactoring and improved paradigms in Waffle v3.
+
+**Removed deprecated APIs from provider**
+
+In :code:`Waffle 3.0.0` we remove deprecated APIs from provider.
```
sz-piotr, line 88: What deprecated apis?

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+
+**Swapped arguments for Fixture**
+
+In :code:`Waffle 3.0.0` we swapped arguments for Fixture, because the provider argument is very rarely used compared to wallets.
+So such implantation should be more convenient for users.
```
sz-piotr, line 93: I guess `implantation` is a word, but you probably meant `implementation`

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+
+*Waffle 2.5.0*
+
+.. code-block:: ts
```
sz-piotr, line 97: Why is the example so verbose? Focus on the core. Define the fixture and call loadFixture on it. No more is needed

docs/source/migration-guides.rst
sz-piotr: Provide the old and new function signature here

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+      const {token, other} = await loadFixture(fixture);
+      await token.transfer(other.address, 7);
+      expect(await token.balanceOf(other.address)).to.equal(7);
+    });
+  });
+
+**Added automatic recognising waffle.json config without cli argument**
+
+Waffle recognises :code:`waffle.json` as default configuration file. If your configuration file is called
+:code:`waffle.json`, it's possible to use just :code:`waffle` to build contracts.
```
sz-piotr, line 154: This is great, but it does not describe what the previous behaviour was if you did not provide the argument to the cli

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,258 @@
...
+      });
+      const wallets = provider.getWallets();
+      expect(wallets.length).to.equal(1);
+      expect(wallets[0].address).to.equal(original.address);
+    });
+  });
+
+**Dropped support for contract.interface**
+
+We dropped support for contract.interface, because it duplicated contract ABI. Now we support just :code:`contract.abi`.
```
sz-piotr, line 227: Several reasons here actually. Originally waffle 1 only had interface, then we replaced `interface` with `abi`, but it was a breaking change. So we added an option to output a legacy version that included `interface`. Unfortunately due to a bug in the code both keys were always outputted. In waffle 3 we dropped interface not only because it is duplicated, but `interface` is a keyword in typescript so you cannot destructure to it.

docs/source/index.rst
```diff
@@ -42,3 +42,4 @@ Versions and ethers compatibility
    configuration.rst
    mock-contract.rst
    ens.rst
+   migration-guides.rst
```
marekkirejczyk, line 45: Can we change order:
   getting-started.rst
   compilation.rst
   basic-testing.rst
   matchers.rst
   fixtures.rst
   mock-contract.rst
   ens.rst
   configuration.rst
   migration-guides.rst

docs/source/migration-guides.rst
```diff
@@ -0,0 +1,252 @@
...
+
+.. code-block:: ts
+
+  await getWallets(provider);
+
+*Waffle 2.3.0*
+
+.. code-block:: ts
+
+  wallets = new MockProvider().getWallets()
```
marekkirejczyk, line 47: `wallets =` not needed

PR #309, opened 2020-06-07
Add waffle-chai usage example

rhlsthrm:
I needed to use these matchers directly without the rest of the Waffle project. The instructions for doing this could be more descriptive, so I added a little extra info.

### DIFF
--- a/waffle-chai/README.md
+++ b/waffle-chai/README.md
@@ -22,6 +22,16 @@ yarn add --dev @ethereum-waffle/chai
 npm install --save-dev @ethereum-waffle/chai
 ```
 
+## Usage
+```ts
+import { expect, use } from "chai";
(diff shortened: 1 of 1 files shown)


waffle-chai/README.md
```diff
@@ -22,6 +22,16 @@ yarn add --dev @ethereum-waffle/chai
...
 ```
 
+## Usage
+```ts
+import { expect, use } from "chai";
+import { waffleChai } from "@ethereum-waffle/chai";
+
+use(chaiAsPromised);
+
+expect(new BigNumber(6)).to.be.gt(0);
```
sz-piotr, line 32: This line will throw an error, because BigNumber is not defined

rhlsthrm: Fixed to be more clear about the BigNumber import.

PR #312, opened 2020-06-10
Use Signer instead of Wallet in ENS module

VladStarostenko:


waffle-provider/test/MockProvider.test.ts
```diff
@@ -55,28 +55,27 @@ describe('INTEGRATION: MockProvider', () => {
     expect(balance.eq(3_141)).to.equal(true);
   });
 
+  let signer: Signer;
+
   describe('ENS', () => {
     before(async () => {
       await provider.setupENS();
+      signer = provider.getSigner(0);
```
marekkirejczyk, line 63: getWallet()

PR #316, opened 2020-06-12
Use Signer instead of Wallet in chai  module

VladStarostenko:


waffle-chai/src/matchers/changeBalance.ts
```diff
@@ -32,10 +32,10 @@ export function supportChangeBalance(Assertion: Chai.AssertionStatic) {
...
 export async function getBalanceChange(
   transactionCallback: () => any,
-  wallet: Wallet
+  signer: Signer
 ) {
-  const balanceBefore = await wallet.getBalance();
+  const balanceBefore = await signer.getBalance();
   await transactionCallback();
-  const balanceAfter = await wallet.getBalance();
+  const balanceAfter = await signer.getBalance();
```
marekkirejczyk, line 39: Hmmm, Are you sure that, this method does exists `signer.getBalance`?

I don't see it in [docs](https://docs.ethers.io/ethers.js/html/api-wallet.html#signer-api) or [source code](https://github.com/ethers-io/ethers.js/blob/master/src.ts/abstract-signer.ts).

PR #321, opened 2020-06-19
Jest Matchers for Smart Contract Development

### LINKED ISSUE #275: Jest support
We should talk about using Waffle with Jest.

In #89, it was discussed that maybe we can import only the matchers from Waffle to be used. What else needs to be done so that Jest users can easily integrate and use Waffle?

Jest is by far the most popular test runner in the JS space (and has been for a while), far exceeding Mocha and Chai:

jest

Source: https://npmcharts.com/compare/mocha,jest,chai

adrianmcli:
Fixes #275 

I have begun work on the Jest matchers based off of the Chai Matchers directory. The rationale for providing Jest matchers is located in #275, but in short:

**Jest is the leading testing framework in the Javascript world**

Jest popularity chart

Chai Matchers (functionality implemented):

- [x] calledOnContract
- [x] bigNumber
- [x] changeBalance
- [x] changeBalances
- [x] emit
- [x] properAddress
- [x] properHex
- [x] properPrivateKey
- [x] reverted
- [x] revertedWith

The following examples are all tested and working as of today (2020-06-20).

# Usage

If you clone the repo and checkout the branch, you can do the following within the monorepo. It's also the expected usage from the end user eventually:

```js
import { waffleJest } from '@ethereum-waffle/jest';

expect.extend(waffleJest);
```

## BigNumbers

```js
expect(await token.balanceOf(wallet.address)).toEqBN(993);
```

## Emitting Events

```js
const tx = await contract.emitOne();

// simple assertion
await expect(tx).toHaveEmitted(contract, 'One');

// assert with args
await expect(tx).toHaveEmittedWith(contract, "One", [
  BigNumber.from(1),
  "One",
  "0x00cfbbaf7ddb3a1476767101c12a0162e241fbad2a0162e2410cfbbaf7162123",
]);
```

## Called on Contract

```js
// simple assertion
await contract.myMethod();

expect('myMethod').toBeCalledOnContract(contract);

// assert with args
await contract.myMethod(1, 2, 3);

expect('myMethod').toBeCalledOnContractWith(contract, [1, 2, 3]);
```

## Revert

```js
// simple revert assertion
await expect(contract.myRevertingFn()).toBeReverted();

// assert with revert reason string
await expect(contract.myRevertingFn()).toBeRevertedWith('Revert reason string');
```

## Balance Changes

```js
const doTransaction = () => sender.sendTransaction({
  to: receiver.address,
  gasPrice: 0,
  value: 200,
});

await expect(doTransaction).toChangeBalances([sender, receiver], ["-200", 200]);
```

## Misc Utils

```js
// assert proper address
expect('0x28FAA621c3348823D6c6548981a19716bcDc740e').toBeProperAddress();

// assert non-proper private key
expect('0x28FAA621c3348823D6c6548981a19716bcDc740').not.toBeProperPrivateKey();

// assert proper hex
expect('0x702').toBeProperHex(2)
```


adrianmcli:
NTS: The event emitting matchers are not really robust. Thing that need to be fixed/tested:

- [ ] Arbitrarily nested data
- [ ] Asserting multiple events on the same transaction
- [ ] Asserting events from multiple contracts on the same transaction

To be honest, we can kill a bunch of those by creating a new matcher where you pass in your transaction hash rather than a promise of the contract mutation.

In either case, I think this is good enough as an `alpha`-level implementation so I will move on to the final `calledOnContract` matchers instead.

adrianmcli:
The preliminary implementation of these matchers is done. It is now basically feature-complete compared with the Waffle Chai matchers. Before I started working on the docs, I wanted to make sure that there is really interest in adopting this into the Waffle repo.

There were also a few changes I had to make due to limitations (and just different conventions) with Jest. I will name some of them here:

## BigNumber Matchers

When overriding existing matchers in Jest, there is no easy way to get access to the underlying matcher that you are overriding. That means we cannot enhance `toEqual` with BigNumber comparisons without breaking the functionality of the native `toEqual` matcher that comes out of the box with Jest.

As a result, I have compromised by adding a `BN` suffix to each matcher instead: `toEqBN`, `toBeGtBN`, `toBeLtBN`, `toBeGteBN`, `toBeLteBN`. It's a little ugly, but hey it's better than nothing. I also haven't gotten around to creating the more wordy aliases either.

## Events API

Rather than passing in the Promise of the transaction, you will pass in the transaction object returned by Ethers.js instead. This is because the extension API with Jest isn't very conducive to custom chaining operators. Therefore, the best way to support asserting multiple events per transaction is to just allow people to make assertions on a transaction object instead.

For example:

```js
// chai matchers
await expect(events.emitOne())
  .to.emit(events, 'One')

// jest matchers
const tx = await events.emitOne();
await expect(tx).toHaveEmitted(events, 'One');
```

## Events Strict Number Comparisons

When asserting Event args with the Chai matchers, a `BigNumber.from(1)` is treated exactly the same as `"1"` or `1`. This is a very understandable convenience feature. But with Jest, it is very handy to use `jest-diff` to display a rich diff of the two arrays of args (expected vs actual).

If we do this, then we need to be more strict with our comparisons. But the benefit of this is a much more user-friendly rich diff in your Jest test output:

Screen Shot 2020-06-20 at 3 45 59 AM

marekkirejczyk:
Hi @adrianmcli,

Looks like you did a lot of awesome work here! We need to allocate a bit more time to review it.
We will try to  start reviewing next week.

Marek

waffle-jest/test/matchers/calledOnContract/calledOnContractWith.test.ts
```diff
@@ -0,0 +1,85 @@
+import {MockProvider} from '@ethereum-waffle/provider';
+import {ContractFactory} from 'ethers';
+import {CALLS_ABI, CALLS_BYTECODE} from '../../contracts/Calls';
+
+async function setup() {
```
sz-piotr, line 5: Deploying the same contract for each test can be done in a fixture, which will speed the tests up significantly. This applies for each of the setup functions that you have that deploy contracts.

waffle-jest/test/matchers/events.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import {EVENTS_ABI, EVENTS_BYTECODE} from '../contracts/Events';
+
+describe('INTEGRATION: Events Emitted', () => {
+  const [wallet] = new MockProvider().getWallets();
+  let factory: ContractFactory;
+  let events: Contract;
+
+  beforeEach(async () => {
+    factory = new ContractFactory(EVENTS_ABI, EVENTS_BYTECODE, wallet);
+    events = await factory.deploy();
```
sz-piotr, line 12: Similarily, this pattern can also be changed to using a fixture, speeding up the tests.

waffle-jest/test/setupJest.ts
```diff
@@ -0,0 +1,4 @@
+import {waffleJest} from '../src';
+
+jest.setTimeout(10000);
+expect.extend(waffleJest);
```
sz-piotr, line 4: It would be nice to have a bit of readme somewhere that specifies how to actually use the matchers in your waffle project

waffle-jest/src/matchers/bigNumber.ts
```diff
@@ -0,0 +1,74 @@
...
+import {Numberish} from '../types';
+
+// NOTE: Jest does not currently support overriding matchers while calling
+// original implementation, therefore we have to name our matchers something
+// different: https://github.com/facebook/jest/issues/6243
+
+export const bigNumberMatchers = {
+  toEqBN(received: Numberish, value: Numberish) {
+    const pass = BigNumber.from(received).eq(value);
+    return pass
```
sz-piotr, line 11: This pattern can be simplified and condensed into:
```ts
return {
  pass,
  message: () => pass
    ? `Expected "${received}" NOT to be greater than ${value}`
    : `Expected "${received}" to be greater than ${value}`
};
```

marekkirejczyk:
Hi @adrianmcli, 

Would love to merge it and publish.  Would you find time to implement @sz-piotr proposals?

adrianmcli:
Sorry, it's been pretty busy for me lately. I'll try to take a look at it this weekend.

waffle-jest/test/matchers/calledOnContract/calledOnContractWith.test.ts
adrianmcli: Jest runs its test suites in parallel, and a custom test environment cannot share a stateful instance that is passed around (see [here](https://github.com/facebook/jest/issues/7184#issuecomment-616037304)).

The only way around this is if you run this mock before you even run your tests and have your tests call out to it.

waffle-jest/src/matchers/bigNumber.ts
adrianmcli: I agree that's more concise, but I'd argue that it's a lot less readable because you are mixing a conditional into an object where one property should correspond to the other.

waffle-jest/test/setupJest.ts
adrianmcli: That is a good idea. I just want to make sure this is something that the project maintainers want before writing it.

waffle-jest/test/matchers/events.test.ts
adrianmcli: See above.

Mrtenz:
Thank you for making this PR. I've tried this PR in my application (using a local version), and it works great. There's a few minor things:

* Some matchers like `toEqBN` and `toBeReverted` don't work when using `await expect().resolves.toEqBN` instead of `await expect().toEqBN`. As far as I know, using `resolves` for promises is the common way for Jest matchers, so it would be nice to see support for it.
* `toHaveEmittedWith` doesn't seem to work. Using `toHaveEmitted` works fine, but when specifying the `expectedArgs`, it fails.

I'd love to see this updated and merged, so we can use this in production.

adrianmcli:
Happy to look into updating the PR further. Can I get a response from @sz-piotr for the comments first?

Also, it might be worth merging in an imperfect implementation than to wait for everything to be fully fleshed out. That being said, I am not a maintainer so this is just a suggestion.

marekkirejczyk:
@Mrtenz: Thanks for feedback!
@adrianmcli: Good point. Let's merge it and publish as experimental and we can fix as we go.
Would you solve conflicts and I can merge & publish tomorrow.

adrianmcli:
@marekkirejczyk sounds good!

adrianmcli:
@marekkirejczyk I have rebased off master to fix the `yarn.lock` conflict and all tests pass.

waffle-jest/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "lint:fix": "eslint --fix '{src,test}/**/*.ts'",
+    "build": "rimraf ./dist && yarn build:esm && yarn build:cjs",
+    "build:esm": "tsc -p tsconfig.build.json --outDir dist/esm --module ES6",
+    "build:cjs": "tsc -p tsconfig.build.json --outDir dist/cjs --declaration false"
+  },
+  "engines": {
+    "node": ">=10.0"
+  },
+  "dependencies": {
+    "@ethereum-waffle/provider": "^3.0.0",
```
marekkirejczyk, line 39: Would be good to remove that dependency, to do so validation in `validateMockProvider` could be relaxed.

waffle-jest/src/matchers/calledOnContract/calledOnContractValidators.ts
```diff
@@ -0,0 +1,44 @@
...
+  ensure(
+    contract instanceof Contract,
+    TypeError,
+    'argument must be a contract'
+  );
+}
+
+export function validateMockProvider(
+  provider: any
+): asserts provider is MockProvider {
```
marekkirejczyk, line 16: We could remove that assertion, lib could work with any provider that has callHistory (e.g. `buidler.dev` provider)

adrianmcli:
@marekkirejczyk removing the assertion in `validateMockProvider` seems to have caused Typescript typecheck to fail:

```
src/matchers/calledOnContract/calledOnContract.ts:16:10 - error TS2339: Property 'callHistory' does not exist on type 'Provider'.

16   const {callHistory} = contract.provider;
            ~~~~~~~~~~~

src/matchers/calledOnContract/calledOnContract.ts:19:6 - error TS7006: Parameter 'call' implicitly has an 'any' type.

19     (call) =>
        ~~~~

src/matchers/calledOnContract/calledOnContractWith.ts:18:10 - error TS2339: Property 'callHistory' does not exist on type 'Provider'.

18   const {callHistory} = contract.provider;
            ~~~~~~~~~~~

src/matchers/calledOnContract/calledOnContractWith.ts:20:6 - error TS7006: Parameter 'call' implicitly has an 'any' type.

20     (call) => call.address === contract.address && call.data === funCallData
        ~~~~
```

I'm not sure how to fix this though. Would be great if you could provide some guidance!

waffle-jest/src/matchers/calledOnContract/calledOnContractValidators.ts
marekkirejczyk: Ahh, now it broke the build I Guess we could define type 
```ts 
interface ProviderWithHistory {
  callHistory ...
}
```
and keep the assertion with it instead of MockProvider

adrianmcli: where would I put this?

waffle-jest/test/matchers/calledOnContract/calledOnContractWith.test.ts
marekkirejczyk: Let's keep it the way it is.

waffle-jest/src/matchers/bigNumber.ts
marekkirejczyk: More than fine for experimental merge

waffle-jest/src/matchers/calledOnContract/calledOnContractValidators.ts
marekkirejczyk: In the same file I guess, it is just for validation purposes to let complier know about the type.
If it becomes problematic, just revert the commit and we can fix in in next version.

marekkirejczyk (approved):

adrianmcli:
hmm, not sure why build is failing. All the tests inside the waffle-jest folder work.

marekkirejczyk:
I downloaded and run test and they pass, error seems to be related to CircleCI out of memory and is a bit random.

adrianmcli:
awesome, thanks for taking a look at this @marekkirejczyk

marekkirejczyk:
Released in [@3.1.0](https://github.com/EthWorks/Waffle/releases/tag/3.1.0)

adrianmcli:
awesome, thanks @marekkirejczyk

rryter:
Hi there.

Thanks a lot for this, works pretty nicely. However my IDE (vs-code) does not understand it:

Screenshot 2020-11-01 at 20 34 17

How do I make vscode recognize the matchers? I tried everything I can think of by now. The only thing that works is to copy and paste the typings file into my project. 

Any help is greatly appreciated.

Mrtenz:
@rryter I was running into the same issue with IntelliJ IDEA. I solved it by adding `types.d.ts` to `files` in `tsconfig.json`:

```json
{
  "files": [
    "node_modules/@ethereum-waffle/jest/src/types.d.ts"
  ]
}
```

adrianmcli:
We should probably add this to the README. Or alternatively, I wonder how we can make the package include the types automatically? @marekkirejczyk

rryter:
> @rryter I was running into the same issue with IntelliJ IDEA. I solved it by adding `types.d.ts` to `files` in `tsconfig.json`:
> 
> ```json
> {
>   "files": [
>     "node_modules/@ethereum-waffle/jest/src/types.d.ts"
>   ]
> }
> ```

I tried that, unfortunately it does not change anything. So maybe I have a "does not work on my machine" problem. 

This is the "affected" repo: https://github.com/rryter/solidity-playground/blob/main/test/ERC734KeyManager.test.ts#L78

Does anybody else use vs-code and could have a quick look? I'll keep digging.

PR #328, opened 2020-07-08
Mocks can now call and staticcall through to contracts

asselstine:
- call will execute a tx on behalf of a mock
- staticcall will call a constant fxn on behalf of a mock


asselstine:
(this is for Waffle 3.0.0)

asselstine:
Changes applied!

One quirk however: Ethers v5 does not seem to support lookup via raw function names.  In his [beta release article](https://blog.ricmoo.com/beta-release-ethers-js-v5-59d0db222d7b) Richard states they would be available if they were unique, but they still weren't.

To amend this, I've added a simple lookup loop so that they can be referenced by name.

sz-piotr:
@marekkirejczyk I am approving this, but not merging, waiting for input on how me merge #329

PR #329, opened 2020-07-08
Mock can call and staticcall through to other contracts

asselstine:
- Mocks can call constant functions
- Mocks can execute transactions

### DIFF
--- a/waffle-mock-contract/src/index.ts
+++ b/waffle-mock-contract/src/index.ts
@@ -37,8 +37,8 @@ export interface MockContract extends Contract {
   mock: {
     [key: string]: Stub;
   };
-  call: Function;
-  staticcall: Function;
+  call (contract: Contract, functionName: string, ...params: any[]): Promise<any>;
+  staticcall (contract: Contract, functionName: string, ...params: any[]): Promise<any>;
 }
 
 export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<MockContract> {
@@ -52,7 +52,7 @@ export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<Mock
   const mockedContract = new Contract(mockContractInstance.address, abi, wallet) as MockContract;
   mockedContract.mock = mock;
 
-  mockedContract.staticcall = async (contract: Contract, functionName: string, ...params: any) => {
+  mockedContract.staticcall = async (contract: Contract, functionName: string, ...params: any[]) => {
     const fn = contract.interface.functions[functionName];
     const data = fn.encode(params);
     let result;
@@ -64,7 +64,7 @@ export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<Mock
     return result;
   };
 
-  mockedContract.call = async (contract: Contract, functionName: string, ...params: any) => {
+  mockedContract.call = async (contract: Contract, functionName: string, ...params: any[]) => {
     const data = contract.interface.functions[functionName].encode(params);
     return mockContractInstance.__waffle__call(contract.address, data);
   };
--- a/waffle-mock-contract/src/index.ts
+++ b/waffle-mock-contract/src/index.ts
@@ -37,6 +37,8 @@ export interface MockContract extends Contract {
   mock: {
     [key: string]: Stub;
   };
+  call: Function,
+  staticcall: Function
 }
 
 export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<MockContract> {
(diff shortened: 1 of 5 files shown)


marekkirejczyk:
@asselstine Interesting idea. 
Would you elaborate on the use case? 

What is the difference between this one and #328?

asselstine:
@marekkirejczyk This is for Waffle 2.5.1, the other is for Waffle 3.0.0

marekkirejczyk:
@asselstine Got it (about versions). Kudos for doing both versions :)

Would you elaborate a bit more why would I like to call another contract through a mock? (ie why not just call directly).
Perhaps you have an example test you would like to show?

asselstine:
It's because the mock contract has privileged execution rights for several functions of the contract under test.

For PoolTogether v3 there are two contracts: the PrizePool and the PrizeStrategy.  The contracts *are* tightly coupled, but the separation allows us to separate the principal from the interest.

You can see it in use [in the PrizeStrategy test](https://github.com/pooltogether/pooltogether-pool-contracts/blob/64bc51d720fa651ebb2a81c0da248d4396d6f7e7/test/PrizeStrategy.test.js#L85).  The test triggers a call through the mock as part of the unit test.

You may notice that the test uses a custom [Doppelganger](https://github.com/pooltogether/pooltogether-pool-contracts/blob/version-3/contracts/test/DoppelgangerWithExec.sol) and [deployMockContract](https://github.com/pooltogether/pooltogether-pool-contracts/blob/version-3/test/helpers/deployMockContract.js) function.  I simply ported them over to the Waffle project for the PRs.

Allowing calls and executes give us *maximum* flexibility when unit testing code.  Additionally, revert messages *are* passed on through the Doppelganger extension, so all errors are visible.

marekkirejczyk:
Make sense to me, let me look at it again with the fresh eye tomorrow.

@sz-piotr Any thoughts?

waffle-mock-contract/src/index.ts
```diff
@@ -37,6 +37,8 @@ export interface MockContract extends Contract {
   mock: {
     [key: string]: Stub;
   };
+  call: Function;
+  staticcall: Function;
```
sz-piotr, line 41: Both call and staticcall have actual signatures:

```ts
export interface MockContract extends Contract {
  mock: {
    [key: string]: Stub;
  };
  call (contract: Contract, functionName: string, ...params: any[]): Promise<any>;
  staticcall (contract: Contract, functionName: string, ...params: any[]): Promise<any>;
}
```

waffle-mock-contract/src/index.ts
```diff
@@ -50,5 +52,22 @@ export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<Mock
   const mockedContract = new Contract(mockContractInstance.address, abi, wallet) as MockContract;
   mockedContract.mock = mock;
 
+  mockedContract.staticcall = async (contract: Contract, functionName: string, ...params: any) => {
```
sz-piotr, line 55:
```
...params: any[]
```

waffle-mock-contract/src/index.ts
```diff
@@ -50,5 +52,22 @@ export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<Mock
...
+    let result;
+    const returnValue = await mockContractInstance.__waffle__staticcall(contract.address, data);
+    result = fn.decode(returnValue);
+    if (result.length === 1) {
+      result = result[0];
+    }
+    return result;
+  };
+
+  mockedContract.call = async (contract: Contract, functionName: string, ...params: any) => {
```
sz-piotr, line 67:
```
...params: any[]
```

waffle-mock-contract/src/index.ts
```diff
@@ -50,5 +52,22 @@ export async function deployMockContract(wallet: Wallet, abi: ABI): Promise<Mock
   const mockedContract = new Contract(mockContractInstance.address, abi, wallet) as MockContract;
   mockedContract.mock = mock;
 
+  mockedContract.staticcall = async (contract: Contract, functionName: string, ...params: any) => {
+    const fn = contract.interface.functions[functionName];
```
sz-piotr, line 56: This also means that you can call it like this: `mockERC20.staticcall(amIRich, 'check')` as opposed to `mockERC20.staticcall(amIRich, 'check()')`, right? We should probably document that fact somehow.

waffle-mock-contract/test/direct.test.ts
```diff
@@ -43,4 +44,21 @@ describe('Mock Contract - Integration (called directly)', () => {
     await expect(mockCounter.add(2)).to.be.revertedWith('Mock revert');
     expect(await mockCounter.add(3)).to.equal(1);
   });
+
+  it('should be able to call to another contract', async () => {
```
sz-piotr, line 48: There should also be a test for specifying just a function name, ignoring the arguments, so `read` instead of `read()`

waffle-mock-contract/src/index.ts
asselstine: Yup; that's the case.  I'll add that to the readme and a test case.

asselstine:
@sz-piotr @marekkirejczyk Ready to go!

sz-piotr:
@marekkirejczyk I am approving this PR, but I am not merging it, because I don't know how we manage the waffle@2 version.

PR #330, opened 2020-07-10
Replace uses of Wallet with Signer

alcuadrado:
Hey @marekkirejczyk @sz-piotr,

I found some issues when using Waffle with Buidler, so I made this PR to fix them.

Some of Waffle's APIs use `Wallet` to type their parameters, but `Signer` could be used instead. This is a problem with Buidler because it sometimes connects to remote networks, and you can't have `Wallet`s in that case.

I changed the fixtures API, and in two chai matches, so that now they receive `Signer`. Luckily this is a backward-compatible change.

Now, the only uses of `Wallet` are left in the `MockProvider`, which I think makes sense.

Do you think we can get this into Waffle `3.0.2`?

### DIFF
--- a/waffle-chai/src/matchers/changeBalance.ts
+++ b/waffle-chai/src/matchers/changeBalance.ts
@@ -1,9 +1,9 @@
-import {Wallet, BigNumber} from 'ethers';
+import {BigNumber, Signer} from 'ethers';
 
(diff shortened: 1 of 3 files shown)


marekkirejczyk:
Merged and releases as 3.0.2 ;)

alcuadrado:
Awesome! Thanks for the quick response.

PR #335, opened 2020-07-19
adding extra BigNumber matchers to the docs

Robsonsjre:
Was looking for this matches at the documentation and they were not there.
Then I saw that these matchers were added in the release (2.4.1 )[#214]


sz-piotr:
Thank you for your contribution <3

PR #336, opened 2020-07-23
Fix source maps

### LINKED ISSUE #281: @ethereum-waffle/provider breaks source maps
Hey 👋

I've received lots of complaints about Buidler lines numbers being off. 

I took some time to debug it, and the root of the problem is that `@ethereum-waffle/chai` loads `@ethereum-waffle/provider`, which in turn loads `ganache-core` and breaks the source maps.

I think I can work around this in Buidler, but I'm not sure how to fix it in Waffle v2.

BW, I think this shows how important my proposal #254 is.

vanruch:
This should fix #281
Looks like line numbers problem was caused by ganache requireing "source-map-support/register".
After removing those requires, line numbers in failed tests are correct, but it needs further testing if it doesn't break anything else

### DIFF
--- a/waffle-provider/package.json
+++ b/waffle-provider/package.json
@@ -35,9 +35,11 @@
     "node": ">=10.0"
   },
   "dependencies": {
+    "@ethereum-waffle/ens": "^3.0.2",
     "ethers": "^5.0.1",
     "ganache-core": "^2.10.2",
-    "@ethereum-waffle/ens": "^3.0.2"
+    "patch-package": "^6.2.2",
+    "postinstall-postinstall": "^2.1.0"
   },
   "resolutions": {
     "web3": "1.2.4"
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "waffle-*"
   ],
   "scripts": {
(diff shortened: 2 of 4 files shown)


marekkirejczyk:
@sz-piotr

waffle-provider/package.json
```diff
@@ -35,9 +35,11 @@
     "node": ">=10.0"
   },
   "dependencies": {
+    "@ethereum-waffle/ens": "^3.0.2",
     "ethers": "^5.0.1",
     "ganache-core": "^2.10.2",
-    "@ethereum-waffle/ens": "^3.0.2"
+    "patch-package": "^6.2.2",
+    "postinstall-postinstall": "^2.1.0"
```
sz-piotr, line 42: why this?

vanruch: https://github.com/ds300/patch-package#why-use-postinstall-postinstall-with-yarn

sz-piotr (approved):

marekkirejczyk:
Released in [3.0.3](https://github.com/EthWorks/Waffle/releases/tag/3.0.3)

PR #349, opened 2020-08-24
Add revert with reason functionality for mock contracts

PaulRBerg:
Not being able to revert with a reason on a mock contract is prohibitively limiting.

Say you have contract A and contract B. The latter depends on the former. Say that the former throws specific errors depending on the state of the properties in the two contracts. If I mock contract A, I have to modify a lot of tests to revert with "Mock revert" instead of my bespoke error messages.

This PR does the following:

+ Adds a new `revertReason` property in the `MockCall` struct
+ Changes the signature of the `__waffle__mockReverts` function, by adding a second parameter `revertReason`
+ Adds a new function `revertWithReason` in the object returned by the `stub` function - which makes my change backwards-compatible!
+ Modifies the tests so that they work with the updated `__waffle__mockReverts` function

### DIFF
--- a/waffle-mock-contract/src/Doppelganger.sol
+++ b/waffle-mock-contract/src/Doppelganger.sol
@@ -5,6 +5,7 @@ contract Doppelganger {
     struct MockCall {
         bool initialized;
         bool reverts;
+        string revertReason;
         bytes returnValue;
     }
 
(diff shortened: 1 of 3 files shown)


marekkirejczyk:
@PaulRBerg Good stuff, merging. Would add a PR with docs?

marekkirejczyk:
Released in [3.0.3](https://github.com/EthWorks/Waffle/releases/tag/3.0.3)

PaulRBerg:
Thanks for merging, Marek!

> Would add a PR with docs?

I could do this, in case you haven't done it in the meantime?

marekkirejczyk:
Released in Waffle 3.1.0. Didn't have a chance to work on docs yet. Would love to get some help :)

PaulRBerg:
Done: https://github.com/EthWorks/Waffle/pull/366.

PR #362, opened 2020-09-21
Fix CI: Split Jest into separate job

SpidiRitsu:


package.json
```diff
@@ -11,7 +11,8 @@
     "postinstall": "patch-package",
     "lint": "wsrun -c lint",
     "build": "wsrun -te -c build",
-    "test": "wsrun -c test"
+    "test": "wsrun -x @ethereum-waffle/jest -c test",
```
marekkirejczyk, line 14: test
test:jest
test:others

marekkirejczyk (approved):

package.json
glinda93: @marekkirejczyk Any plan to bring back jest?

marekkirejczyk: Generally we will deprecate JEST support, as it didn't get traction.
Current version supports JEST.

PR #373, opened 2020-09-28
Add missing spread operator

### LINKED ISSUE #340: Bug in mock contract call
For Waffle 3.0.0:

This [line in index.ts](https://github.com/EthWorks/Waffle/blob/79b68298dff18a7770bdc013bb98da7a1df8ea63/waffle-mock-contract/src/index.ts#L87) currently reads:

```
const tx = await contract.populateTransaction[functionName](params);
```

However for Ethers V5 it needs to spread the params like so:

```
const tx = await contract.populateTransaction[functionName](...params);
```

duckception:
Adds missing spread syntax for the parameters for populateTransaction call as specified in [ethers v5 documentation](https://docs.ethers.io/v5/api/contract/contract/#contract-populateTransaction).

Closes #340

### DIFF
--- a/waffle-mock-contract/src/index.ts
+++ b/waffle-mock-contract/src/index.ts
@@ -85,7 +85,7 @@ export async function deployMockContract(signer: Signer, abi: ABI): Promise<Mock
   };
 
   mockedContract.call = async (contract: Contract, functionName: string, ...params: any[]) => {
-    const tx = await contract.populateTransaction[functionName](params);
+    const tx = await contract.populateTransaction[functionName](...params);
     const data = tx.data;
     return mockContractInstance.__waffle__call(contract.address, data);
   };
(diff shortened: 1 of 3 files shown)


waffle-mock-contract/src/index.ts
```diff
@@ -85,7 +85,7 @@ export async function deployMockContract(signer: Signer, abi: ABI): Promise<Mock
   };
 
   mockedContract.call = async (contract: Contract, functionName: string, ...params: any[]) => {
-    const tx = await contract.populateTransaction[functionName](params);
+    const tx = await contract.populateTransaction[functionName](...params);
```
marekkirejczyk, line 88: We should add a test that fails in the old scenario and succeeds after the fix.

marekkirejczyk (changes requested):

marekkirejczyk (approved):

marekkirejczyk (approved):

PR #391, opened 2020-11-15
Make solcjs compilation work with versions >=0.7.2

msieczko:
For calculating solcjs binaries download paths we use a list of compiler versions fetched from [this github page](https://ethereum.github.io/solc-bin/bin/list.json). It seems the repository that publishes it [run out of disc space](https://github.com/ethereum/solc-bin/runs/1179669192) and the list has not been updated since September 26.

I changed the list url to point to the current version of the file taken directly from github repository.

### DIFF
--- a/waffle-compiler/src/compileSolcjs.ts
+++ b/waffle-compiler/src/compileSolcjs.ts
@@ -50,7 +50,7 @@ async function resolveSemverVersion(version: string) {
   return item.substring('soljson-'.length, item.length - '.js'.length);
 }
 
-const VERSION_LIST_URL = 'https://ethereum.github.io/solc-bin/bin/list.json';
+const VERSION_LIST_URL = 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json';
 let cache: any = undefined;
 
 async function fetchReleases() {


marekkirejczyk (approved):

waffle-compiler/src/compileSolcjs.ts
```diff
@@ -50,7 +50,7 @@ async function resolveSemverVersion(version: string) {
   return item.substring('soljson-'.length, item.length - '.js'.length);
 }
 
-const VERSION_LIST_URL = 'https://ethereum.github.io/solc-bin/bin/list.json';
+const VERSION_LIST_URL = 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json';
```
axic, line 53: The preferred URL is https://solc-bin.ethereum.org/bin/list.json I believe. @cameel could you confirm? We should document this as well.

cameel: That's right. Our repo exceeded GH pages size limit and we switched to hosting the binaries elsewhere. Unfortunately we don't control the ethereum.github.io domain and could not provide a redirect. The transition was seamless for anyone using `solc-bin.ethereum.org` which we do control.

```suggestion
const VERSION_LIST_URL = 'https://solc-bin.ethereum.org/bin/list.json';
```

I would also recommend switching from `bin/` to `emscripten-wasm32/` directory with a fallback to `emscripten-asmjs/` if a version is not available there. `emscripten-wasm32/` contains wasm builds which offer much better performance than asm.js. Files in `bin/` match releases on github and these are wasm-based only starting with 0.6.2 while `emscripten-wasm32/` has rebuilt binaries for older versions too.

> We should document this as well.

I'm on that. I have a half-finished documentation update for this already but I got distracted by other tasks. I'll finish that and push a PR today or tomorrow.

cameel: Also note that fetching files from the git repo via `raw.githubusercontent.com` is not the best idea because some files in it are symlinks. They are rendered into proper redirects or copies on GH pages and on `solc-bin.ethereum.org` but github will give you a short text file with a relative path instead. `list.json` is safe but you'll get into that problem if you try that with some of the binaries (e.g. those with shortened commit hash that we symlinked under proper names when we discovered that they get shortened).

PR #395, opened 2020-11-19
Add option to select compiler path in case of native solc

axic:
This reuses the config.compilerVersion field.


waffle-compiler/test/compiler/wrappers/compileNative.ts
```diff
@@ -19,6 +19,16 @@ describe('UNIT: compileNative', () => {
     expect(actualCommand).to.match(new RegExp(expectedCommand));
   });
 
+  it('buildCommand with custom compiler', async () => {
```
axic, line 22: I couldn't actually execute the test suite as it fails with various issues on macOS. One being the lack of docker.

waffle-compiler/src/compileNativeSolc.ts
```diff
@@ -14,7 +14,7 @@ export function compileNativeSolc(config: Config) {
 }
 
 export function createBuildCommand(config: Config) {
-  const command = 'solc';
+  const command = config.compilerVersion || 'solc';
```
axic, line 17: Not sure if reusing this field is a good idea, but this was the least invasive change.

axic:
I think the test failures are not introduced by this PR.

waffle-compiler/src/compileNativeSolc.ts
```diff
@@ -14,7 +14,7 @@ export function compileNativeSolc(config: Config) {
 }
 
 export function createBuildCommand(config: Config) {
-  const command = 'solc';
+  const command = (config.compilerVersion === 'default') ? 'solc' : config.compilerVersion;
```
axic, line 17: I'm not a big typescript guru so not sure why this still defaults to `default` somehow, according to the tests.

rzadp:
I think we want to keep the current way of building the `solc` compilation command, and if the user wants to use different version than the default, he or she can alias the `solc` to a specific version in the shell.

LMK if that does not sounds like a viable solution.

PR #396, opened 2020-11-20
Fetch more efficient wasm32 compiler binaries for versions < 0.6.2

msieczko:
This PR introduces changes suggested by @cameel in https://github.com/EthWorks/Waffle/pull/391#discussion_r526894609.

Leaving it as draft because the CI still fails on one environment for some reason.

### DIFF
--- a/waffle-compiler/src/compileSolcjs.ts
+++ b/waffle-compiler/src/compileSolcjs.ts
@@ -50,7 +50,7 @@ async function resolveSemverVersion(version: string) {
   return item.substring('soljson-'.length, item.length - '.js'.length);
 }
 
-const VERSION_LIST_URL = 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json';
+const VERSION_LIST_URL = 'https://solc-bin.ethereum.org/bin/list.json';
 let cache: any = undefined;
 
 async function fetchReleases() {
--- a/waffle-compiler/src/compileSolcjs.ts
+++ b/waffle-compiler/src/compileSolcjs.ts
@@ -9,8 +9,10 @@ import {findImports} from './findImports';
 import mkdirp from 'mkdirp';
 import fs from 'fs';
 import https from 'https';
+import type http from 'http';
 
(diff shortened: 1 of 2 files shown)


waffle-compiler/src/compileSolcjs.ts
```diff
@@ -50,7 +52,7 @@ async function resolveSemverVersion(version: string) {
   return item.substring('soljson-'.length, item.length - '.js'.length);
 }
 
-const VERSION_LIST_URL = 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json';
+const VERSION_LIST_URL = 'https://solc-bin.ethereum.org/bin/list.json';
```
cameel, line 55: Please note that every directory has its own list. If you're getting files from `emscripten-wasm32/` and `emscripten-asmjs/` now, it would be safer to fetch `emscripten-wasm32/list.json` and `emscripten-asmjs/list.json`. `bin/list.json` will work as long as all you need is the name because we're adding new releases to both `emscripten-wasm32/` and `bin/`  but would be a problem if you wanted to e.g. verify the keccak256 checksum listed in the JSON file (which will be different if the binary in `bin/` is the asm.js one).

By the way, here's a draft of the documentation update for solc-bin: https://github.com/ethereum/solidity/pull/10352. Any feedback is welcome in case anything is still missing.

msieczko: @cameel thanks for your valuable input! For now we're just using the `list.json` file to get the release versions and their commit hashes. I'll consider switching to fetching both of the lists you mentioned when I revisit the PR.

marekkirejczyk:
@msieczko  Would you like to finalise and solve conflicts, so that we can merge?

msieczko:
The changes from this PR make our CI fail on Node v10 for a reason mentioned here: https://github.com/EthWorks/Waffle/pull/397#issuecomment-733057618. I don't think it's worth debugging and I'd close this PR. The most important thing which is correcting the URL we fetch binaries from is already merged in https://github.com/EthWorks/Waffle/pull/397.

PR #397, opened 2020-11-24
Correct version list url

sz-piotr:
Changes the version list url to `https://solc-bin.ethereum.org/bin/list.json`.
The change in mocha.opts is to prevent mocha from loading the cached versions of solc saved as js files.

@cameel I wanted to use the emscripten-wasm32 builds, but found that for some reason they aren't faster and using them made the tests timeout. I don't have more time to investigate the issue so we are sticking with `bin` for now.

Edit: the issue seems to only be related to node 10 #396. I suspect we will merge this PR and add support for wasm at a later date

### DIFF
--- a/waffle-compiler/src/compileSolcjs.ts
+++ b/waffle-compiler/src/compileSolcjs.ts
@@ -50,7 +50,7 @@ async function resolveSemverVersion(version: string) {
   return item.substring('soljson-'.length, item.length - '.js'.length);
 }
 
-const VERSION_LIST_URL = 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json';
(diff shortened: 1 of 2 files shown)


marekkirejczyk (approved):

cameel:
Yeah, this is a weird problem that shows up only on older nodejs that we haven't been able to solve so far: https://github.com/ethereum/solidity/issues/9297. On node 12 and above there's no delay and node 10 will be EOL in a few months so probably not worth focusing on but if you have some idea what might be causing this we could try to prepare a fix.

marekkirejczyk:
@cameel  Check freshly released version [3.2.1](https://github.com/EthWorks/Waffle/releases/tag/3.2.1)

PR #401, opened 2020-11-26
Update dev dependencies, replace node 11 with 14

sz-piotr:
Node 11 was dopped because typescript-eslint no longer supports it, as well as the fact that it reached its end of life


marekkirejczyk (approved):

marekkirejczyk:
Released in [3.2.2](https://github.com/EthWorks/Waffle/releases/tag/3.2.2)

PR #406, opened 2020-12-04
Fix missing single quote in README.md

nijynot:
The README.md was missing a single quote in one of the chai matchers.


marekkirejczyk (approved):

marekkirejczyk:
Good catch! Thanks @nijynot.

Issue #87, opened 2019-03-04
Events: Support indexed array attributes in tests


svenstucki:
The current `to.emit().withArgs()` matcher does not support indexed (variable length) arrays. Indexed attributes must be fixed size (32 bytes), so solidity stores a hash of the array content. This is not checked for and the comparison in a test fails.

Example contract:

```
pragma solidity ^0.5.4;

contract Test {
  event Event (bytes a, bytes indexed b);

  function emitEvent() public {
    bytes memory test = abi.encode(0x01);
    emit Event(test, test);
  }
}
```

Example test case:
```
import chai from 'chai';
import { createMockProvider, deployContract, getWallets, solidity } from 'ethereum-waffle';

import Test from '../build/Test';

chai.use(solidity);
const { expect } = chai;

describe('Test', () => {
  const provider = createMockProvider();
  const [wallet] = getWallets(provider);

  const packed = '0x0000000000000000000000000000000000000000000000000000000000000001';

  it('Should emit', async () => {
    const t = await deployContract(wallet, Test, []);

    await expect(t.emitEvent())
      .to.emit(t, 'Event')
      .withArgs(packed, packed);
  });
});
```

The `actualArgs.values` in matchers.ts line 113 contains the following:
```
     '0': '0x0000000000000000000000000000000000000000000000000000000000000001',
     '1':
      _Indexed {
        hash: '0xb10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6' },
```

Which is not unexpected, but the comparison fails for the indexed argument. When passing an object in the test (i.e. `{ hash: '0xb10...' }` instead of the `0x0..01` string), the comparison fails as well.

It would be great to have a simple way to test indexed attributes.

sz-piotr:
We would gladly merge a PR that adds this functionality!

zgorizzo69:
@sz-piotr I have a simple PR for that. Can you give me the creds to push a branch ?

sz-piotr:
@zgorizzo69 You can fork the repository and open a PR this way :)

wachulski:
Shouldn't this one be closed? @sz-piotr ?

sz-piotr:
Thanks @wachulski
Closing.
Please note that I am no longer a maintainer and ping @marekkirejczyk in the future.

wachulski:
Yup, I noticed that @sz-piotr and hence created #467 Thanks.

PR #413, opened 2020-12-14
feat(event): add assertion for indexed dynamic size parameters

zgorizzo69:
The current to.emit().withArgs() matcher does not support indexed (variable length) arrays.  
To fix that we check if the argument is indexed and if there is a `hash` property
If it exist we assert equal based on that `hash` property 
linked to #87


zgorizzo69:
@sz-piotr I would greatly appreciate your input on this :pray:

waffle-chai/test/matchers/events.test.ts
```diff
@@ -27,7 +27,9 @@ describe('INTEGRATION: Events', () => {
   });
 
   it('Emit two: success', async () => {
-    await expect(events.emitTwo()).to.emit(events, 'Two');
+    await expect(
```
sz-piotr, line 30: Please format as:
```ts
await expect(events.emitTwo())
  .to.emit(events, 'Two')
  .withArgs(2, 'Two');
```

waffle-chai/test/contracts/Events.ts
```diff
@@ -16,8 +17,16 @@ export const EVENTS_SOURCE = `
           emit One(1, "DifferentKindOfOne", 0x0000000000000000000000000000000000000000000000000000000000000001);
       }
 
+      function emitIndex() public {
+        emit Index("Three",
```
sz-piotr, line 21: Add a newline after `(`

waffle-chai/test/matchers/events.test.ts
```diff
@@ -39,6 +41,15 @@ describe('INTEGRATION: Events', () => {
     );
   });
 
+  it('Emit index: success', async () => {
+    const bytes = ethers.utils.hexlify(ethers.utils.toUtf8Bytes('Three'));
+    const hash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes('Three'));
+    await expect(
```
sz-piotr, line 47: Please format as:
```ts
await expect(events.emitIndex())
  .to.emit(events, 'Index')
  .withArgs(
    hash,
    'Three',
    bytes,
    hash,
   '0x00cfbbaf7ddb3a1476767101c12a0162e241fbad2a0162e2410cfbbaf7162123'
  )
```

sz-piotr: Wouldn't it make more sense for the expectation to look like this?
Under the hood the matcher would hash the required arguments. In this way the user doesn't have to figure out the hashing themselves. Alternatively we can support both - if a hash is provided it is checked directly, otherwise a hash is computed on the fly.
```ts
  .withArgs(
    "Three",
    "Three",
    bytes,
    bytes,
    '0x00cFBbaF7DDB3a1476767101c12a0162e241fbAD2a0162e2410cFBbaF7162123'
);
```

sz-piotr (changes requested):

sz-piotr:
@zgorizzo69 Thanks for submitting the PR. I have requested some changes and I also welcome your opinion on them.

waffle-chai/test/matchers/events.test.ts
zgorizzo69: yes as you can see in the commit that is more or less what i did in the previous commit I will update it. What prettier conf / eslint rules are you using exactly ?

zgorizzo69: @sz-piotr done

marekkirejczyk (approved):

marekkirejczyk (approved):

marekkirejczyk:
Looks good. Would you do a bit of docs before release?

zgorizzo69:
@marekkirejczyk can you check [here](https://github.com/EthWorks/Waffle/pull/426) if it follow your standards ? If it is too long we could only leave the `note` block

marekkirejczyk:
Released in [3.2.2](https://github.com/EthWorks/Waffle/releases/tag/3.2.2)

PR #442, opened 2021-02-26
Extend revertedWith functionality

### LINKED ISSUE #400: revertedWith succeeds despite only partial match

malyfko:
Closes #400


marekkirejczyk (approved):

skywinder:
Hello, maintainers, **this Error still persist in `3.4.0`**

cc @marekkirejczyk 

False-negative scenario if I remove the beginning or ending part of `reveredWith` parameter!

### True-negative case (original string):

CleanShot 2021-08-19 at 18 13 41@2x


### False-negative scenario (it pass, but should fail)

CleanShot 2021-08-19 at 18 13 19@2x

PR #457, opened 2021-03-03
Align getting-started.rst with the basic example

wachulski:
Adding missing information contained in `examples/basic`.

Saves time especially for beginners that start with the docs and assumes the page contains a complete list of steps to do.


marekkirejczyk (approved):

marekkirejczyk:
This is a good point, but note, that it is only require if you use typescript in your project. We should add a note about it.

marekkirejczyk:
Good catch, it is only the case if one is using TypeScript, we should add such a notes.

docs/source/getting-started.rst
```diff
@@ -165,6 +165,8 @@ Run:
 
       npm install --save-dev mocha chai
 
+You may also need to add :code:`ts-node` and :code:`typescript` if not installed globally.
```
marekkirejczyk, line 168: Should be something more along these lines:
If you are using Typescirpt don't forget to add `ts-node` to your `devDependencies`

docs/source/getting-started.rst
```diff
@@ -235,6 +237,15 @@ Update your :code:`package.json` file to include:
     }
   }
 
+Add :code:`mocha.opts` file in your test folder:
```
marekkirejczyk, line 240: If you are using TypeScript add ....

docs/source/getting-started.rst
wachulski: Fixed. My intent here was to make it self-contained for a newcomer (i.e. going through it step by step literally should produce the described results). Even sth may be obvious to the more experienced.

docs/source/getting-started.rst
wachulski: Done

marekkirejczyk (approved):

marekkirejczyk (approved):

Issue #382, opened 2020-10-14
Ethers caching block number causes problems with fixtures


vanruch:
When there are some tests with fixtures, the block numbers are returned incorrectly by `provider.getBlockNumber()`.
It's caused by Ethers [caching](https://github.com/ethers-io/ethers.js/blob/5e0e3de52ecead2304ae8db68b7e67ceed9197a2/packages/providers/src.ts/base-provider.ts#L616) last block number.

**Steps to reproduce**
```{js}
const fixture = async (wallets, _provider) => {
  [owner] = wallets
  provider = _provider
}

beforeEach(async () => {
  await loadFixture(fixture)
})

it('0', async () => {
  console.log(await provider.getBlockNumber())
  await owner.sendTransaction({ value: '0x1', to: AddressZero })
  await provider.getBlockNumber()
})

it('1', async () => {
  console.log(await provider.getBlockNumber())
})
```
It prints `0` and `1`, while should print `0` for both tests

wachulski:
I need some help for to fix that properly. Since it's marked 'Good first issue' it shouldn't spawn such concerns I have in #466 . Most likely I'm missing some pieces in the provider facilities. Could you guide me through @vanruch ?

vanruch:
Thanks for looking into it @wachulski 
To be fair, I can't think of any other solution other than your's. Ethers for example sets _maxInternalBlockNumber to -1024 on every network change https://github.com/ethers-io/ethers.js/blob/5e0e3de52ecead2304ae8db68b7e67ceed9197a2/packages/providers/src.ts/base-provider.ts#L779. 
Maybe we should even copy all this code for resetting block number guards and caches?

wachulski:
That's very fragile IMHO. That's internals:
> // Reset all internal block number guards and caches

If we do so, there is significant risk it will interfere with some other state we are not aware yet and hit fixture creation/load with dirty provider state delivered.
Now or in the future (as base provider authors may change their implementation in minor version bumps, no?)

> Ethers for example sets _maxInternalBlockNumber to -1024 on every network change https://github.com/ethers-io/ethers.js/blob/5e0e3de52ecead2304ae8db68b7e67ceed9197a2/packages/providers/src.ts/base-provider.ts#L779

Yeah, but it's internals they do have control over (changes, releasing without major version bump, etc.)
[...]

PR #466, opened 2021-03-15
Rough fix proposal as foundation for discussion

wachulski:
As I wrote:
// revert provider internal state (NOTE!!! This is fragile)
snapshot.provider._maxInternalBlockNumber = 0;

I believe that's not an option from the good quality standpoint.

I also see that the base provider resets its internal cache/state on network
changes. But do you know if/how I can use it straightforward in the mock provider? Even if yes, doesn't it contradict the fundamental design of Waffle (to have fixtures reset quickly and have good perf.; if such networks resets slow even in mocked env, so the perf deteriorates right)?


wachulski:
Closes #382

wachulski:
Got stale, closing

PR #478, opened 2021-03-24
chore: trim example deps

gakonst:
As title, makes examples use the latest waffle, trimming some more redundant deps there.


gakonst:
@marekkirejczyk the 12.x node tests seem to segfault, unclear why? In any case this PR also should be easy to merge. Not sure if you want to bump the version in the examples, probably not?

marekkirejczyk:
❤️ Great work, thanks!

gakonst:
There were patches made in the `dependencies` for `waffle-{compiler, e2e, jest, mock-contract, chai}`, so it seems like we do need to publish a new patch version for them, no?

Re examples: I updated the versions so that I can update the examples' versions, since they didn't seem to be updating with changesets.

Do you want me to make any change in the PR (I think if you know specifically what needs to change

rzadp:
@gakonst Thanks for the PR, which now got outdated.
In the meantime we did some overlaping steps:

- old `ganache-core` has been updated
- `pnpm` is used, which allows us to have always up-to-date usedapp dependencies in the example app using `workspace:*` references

PR #573, opened 2021-09-27
Upgrade set-value, yargs-parser

jyap808:
Fixes CVE-2021-23440, CVE-2020-7608


rzadp:
Thanks for the PR. Closing as no longer relevant, `set-value` is not used and `yargs-parser` is up-to-date.

Issue #583, opened 2021-10-05
this throws when the event parameter is undefined


noobshow:
https://github.com/EthWorks/Waffle/blob/e93d80233394d1feae2acfdbf6f4c06abbf8ee86/waffle-chai/src/matchers/emit.ts#L64

PR #584, opened 2021-10-05
🧛‍♂️ Check for undefined event parameter

noobshow:
Error thrown when event returns nullish/undefined parameter like empty string or empty array:
`     TypeError: Cannot read property 'length' of undefined
`
This pull request closes https://github.com/EthWorks/Waffle/issues/583

### DIFF
--- a/waffle-chai/src/matchers/emit.ts
+++ b/waffle-chai/src/matchers/emit.ts
@@ -61,7 +61,7 @@ export function supportEmit(Assertion: Chai.AssertionStatic) {
       actualArgs.length
     );
     for (let index = 0; index < expectedArgs.length; index++) {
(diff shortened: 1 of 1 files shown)


waffle-chai/src/matchers/emit.ts
```diff
@@ -61,7 +61,7 @@ export function supportEmit(Assertion: Chai.AssertionStatic) {
       actualArgs.length
     );
     for (let index = 0; index < expectedArgs.length; index++) {
-      if (expectedArgs[index].length !== undefined && typeof expectedArgs[index] !== 'string') {
+      if (expectedArgs[index] !== undefined && expectedArgs[index].length !== undefined && typeof expectedArgs[index] !== 'string') {
```
rzadp, line 64:
```suggestion
      if (expectedArgs[index]?.length !== undefined && typeof expectedArgs[index] !== 'string') {
```

rzadp (approved):

Issue #580, opened 2021-09-28
Incompatibility with latest ethers version


tuler:
The latest version of ethers includes support for London fork by extending the Provider interface with an additional method.
This breaks waffle's MockProvider class.

```
Argument of type 'MockProvider' is not assignable to parameter of type 'Provider'.
  Property 'getFeeData' is missing in type 'MockProvider' but required in type 'Provider'
```

D4nte:
Yes, same issue when using the fixtures with latest `ethers`:

```
| test/1.votingContract.test.ts(94,42): error TS2345: Argument of type '([alice, firstAddress, secondAddress, noTokensAddress]: any[], provider: JsonRpcProvider) => Promise<{ contract: Contract; secondContract: Contract; alice: any; firstAddress: any; secondAddress: any; provider: JsonRpcProvider; noTokensAddress: any; }>' is not assignable to parameter of type 'Fixture<{ contract: Contract; secondContract: Contract; alice: any; firstAddress: any; secondAddress: any; provider: JsonRpcProvider; noTokensAddress: any; }>'.
 |   Types of parameters 'provider' and 'provider' are incompatible.
```

rzadp:
This is now sorted out in Waffle 4.
Internally we updated deprecated `ganache-core` to `ganache` which supports London hardfork.

PR #603, opened 2021-12-24
Bump Ethers

D4nte:
Solves https://github.com/EthWorks/Waffle/issues/580.


D4nte:
After more analysis I think ethers introduced a breaking change in a minor version bump (5.4 to 5.5) which is why there are so many issues around this dependency.

Which means I am not sure whether this change warrants a patch or major version bump of waffle.

Szymx95:
@marekkirejczyk

PR #645, opened 2022-03-16
Fixed problems with getting started tutorial

samlaf:
1. openzeppelin/contracts installs contracts that use solidity version `^0.8.0`, and are not compatible with the compiler `0.6.2` used in the examples here. I changed it to install `@openzeppelin/contracts@3.4.2` instead.
2. `mocha.opts` is no longer used (started mocha version 8 I believe). I changed the config file to the new `.mocharc.json`
3. The instruction `Lastly, instead of using mocha in your Node scripts within your package.json, replace it with ts-node.` was confusing to me as I didn't need it... perhaps it's meant for one of the next tutorials? I removed it and added `npm i -D @types/mocha` instead, which is needed in typescript to lint the `describe`, `beforeEach`, etc.

Closes https://github.com/TrueFiEng/Waffle/issues/539


rzadp:
@samlaf Thank you for the PR.

1. I have updated the compiler to make it work with solidity `0.8.0`
2. Good catch, thanks
3. Agree, added some edits as well

rzadp (approved):

PR #676, opened 2022-03-25
Jest v3.2.2

yivlad:


mdnorman:
@yivlad why was waffle-jest deprecated? I and others are using it, and it works great.

PR #681, opened 2022-03-29
🪟 Preserve old behaviour of failing on sending reverting transaction

rzadp:


waffle-provider/src/revertString.ts
```diff
@@ -66,24 +85,29 @@ export const injectRevertString = (provider: Provider): Provider => {
...
           })();
+        } else if (method === 'eth_sendRawTransaction') {
+          /**
+           * Because we have overriden the gas estimation not to be failing on reverts,
+           * we add a wait during transaction sending to retain original behaviour of
+           * having an exception when sending a failing transaction.
+           */
+          return (async () => {
+            const transactionHash = await originalResult;
+            const etherProvider = new providers.Web3Provider(provider as any);
```
dmaretskyi, line 96: extract

yivlad (approved):

PR #683, opened 2022-04-01
🏄  Migration and fixes after introducing new Ganache

rzadp:
- Run test on a combination of different Ganache provider options
- Migration guide about using `gasPrice` and wallet mnemonic
- Update tests to make them work on London hardfork
- More flexible gas estimation call back uses block gas limit instead of arbitrary value


mj426382 (approved):
Everything looks correct

PR #685, opened 2022-04-06
🪖 Waffle hardhat testing

yivlad:


waffle-provider/test/MockProvider.test.ts
```diff
@@ -5,7 +5,7 @@ import {deployToken} from './BasicToken';
 import {describeMockProviderCases} from './MockProviderCases';
 
 describeMockProviderCases('INTEGRATION: MockProvider', (provider) => {
-  it('returns wallets', async () => {
+  it.only('returns wallets', async () => {
```
rzadp, line 8:
```suggestion
  it('returns wallets', async () => {
```
```suggestion
  it.only('returns wallets', async () => {
```

rzadp, line 8:
```suggestion
  it('returns wallets', async () => {
```
```suggestion
  it.only('returns wallets', async () => {
```

rzadp: Created https://github.com/TrueFiEng/useDApp/runs/5882107912?check_suite_focus=true

rzadp (approved):

PR #692, opened 2022-04-12
Chaining emit matchers

yivlad:


rzadp (commented):
Can you also add some more test cases for not-awaited `tx`, like:

```ts
it('Both emitted and caught', async () => {
      await expect(events.emitBoth())
        .to.emit(events, 'One')
        .to.emit(events, 'Two');
    });
```

dmaretskyi (approved):

PR #726, opened 2022-05-10
🎥  Reverted with custom error test args

yivlad:


waffle-hardhat/test/reverted.test.ts
```diff
@@ -24,12 +24,111 @@ describe('INTEGRATION: Matchers: revertedWith', () => {
...
+  const deploy = async () => {
+    const wallets = provider.getWallets();
     const wallet = wallets[0];
     const factory = new ContractFactory(CustomError.abi, CustomError.bytecode, wallet);
     const matchers = await factory.deploy();
-    await expect(matchers.doRevertWithCustomError()).to.be.revertedWith('CustomError(0)');
+    return matchers;
+  };
+
+  describe('Custom errors', () => {
```
rzadp, line 35: Add a comment that these test cases are not in common `revertedWithTest` because they are exclusive to hardhat. Or are they not? maybe we will make them work for L2 as well? maybe we would like something like this:

```ts
revertedWithTest(provider, {includeCustomErrors: true});
```

rzadp (approved):

PR #727, opened 2022-05-12
🧮 Change ether balance calculate fees

yivlad:


waffle-provider/src/MockProvider.ts
```diff
@@ -83,4 +83,9 @@ export class MockProvider extends providers.Web3Provider {
     this.network.ensAddress = ens.ens.address;
     this._ens = ens;
   }
+
+  async getTransactionFee(txHash: string): Promise<BigNumber> {
```
rzadp, line 87: Not sure about adding this to the public API, this will become part of our public interface without need.
How about moving those two lines to the tests where it's needed as an util function?

PR #755, opened 2022-07-13
👷 Add script for supporting calledOnContract matcher in hardhat

yivlad:


waffle-hardhat/src/inject-call-history.ts
```diff
@@ -0,0 +1,61 @@
+import {waffle} from 'hardhat';
+import {RecordedCall} from 'ethereum-waffle';
```
rzadp, line 2: type, devdep

waffle-hardhat/package.json
```diff
@@ -18,15 +18,17 @@
...
     "node": ">=10.0"
   },
+  "dependencies": {
+    "ethers": "5.6.2",
+    "hardhat": "^2.9.2"
+  },
   "devDependencies": {
     "@ethereum-waffle/chai": "workspace:*",
     "@nomiclabs/hardhat-ethers": "^2.0.3",
     "@nomiclabs/hardhat-waffle": "^2.0.3",
```
rzadp, line 28: `@nomiclabs/hardhat-waffle` and `hardhat` - can only one of those be used?

waffle-hardhat/src/inject-call-history.ts
```diff
@@ -0,0 +1,61 @@
+import {waffle} from 'hardhat';
+import type {RecordedCall} from 'ethereum-waffle';
+import {utils} from 'ethers';
+
+class CallHistory {
```
rzadp, line 5: How about defining an `ICallHistory` interface in waffle-provider, importing here (as type from dev dep), so that the interface is kept in sync?

waffle-hardhat/src/inject-call-history.ts
```diff
@@ -0,0 +1,61 @@
...
+      ._node
+      ._vmTracer
+      ._vm
+      .off('beforeMessage', beforeMessageListener);
+  }
+  beforeMessageListener = (message: any) => {
+    callHistory.addUniqueCall(toRecordedCall(message));
+  };
+  const provider: any = waffle.provider;
+  provider.callHistory = callHistory.recordedCalls;
```
rzadp, line 52: Does it still work properly when we do `clearAll`? The array reference changes because we create a new one - `this.recordedCalls = []`

waffle-hardhat/package.json
yivlad: hardhat is peer dep of @nomiclabs/hardhat-waffle

waffle-hardhat/package.json
```diff
@@ -18,15 +18,17 @@
   "engines": {
     "node": ">=10.0"
   },
+  "dependencies": {
+    "ethers": "5.6.2",
+    "hardhat": "^2.9.2"
+  },
   "devDependencies": {
     "@ethereum-waffle/chai": "workspace:*",
     "@nomiclabs/hardhat-ethers": "^2.0.3",
```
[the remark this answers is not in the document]
yivlad: Same as above

PR #765, opened 2022-07-26
Add contract function hints to the MockContract.mock object

ackvf:
This change is backwards compatible as it uses an optional generic, that is by default set to `BaseContract`, which contains a mapped object `functions: { [ name: string ]: ... };` that provides the same namespace.

I've been using this already with this ambient type declaration
```
declare type BaseContract = import('ethers').BaseContract
declare type _MockContract = import('ethereum-waffle').MockContract
declare type Stub = import('ethereum-waffle').Stub
interface MockContract<T extends BaseContract = BaseContract> extends _MockContract {
  mock: {
    [key in keyof T['functions']]: Stub
  }
}
```

image

### DIFF
--- a/.changeset/two-rice-teach.md
+++ b/.changeset/two-rice-teach.md
@@ -1,0 +1,5 @@
+---
+"@ethereum-waffle/mock-contract": patch
+---
+
+Add contract function hints to the `MockContract.mock` object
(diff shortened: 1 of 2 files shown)


ackvf:
Dang, I put that commit message there by mistake. I thought it was the PR text. You can amend my commit and remove it if you want.

rzadp:
@ackvf We have some build errors, did you take a look at those?

ackvf:
@rzadp I didn't see this before. I will look at it.

PR #784, opened 2022-08-09
🧅 Allow special characters in revertedWith regex

iconofsin:
Update the custom error regex to accommodate Solidity's definition of a valid identifier: https://docs.soliditylang.org/en/v0.8.16/grammar.html#a4.SolidityLexer.Identifier

Motivation: allow matching custom errors with names containing $ and _. Currently, if a custom error's name contains either of these characters, the matcher will return "" instead of the error name.


rzadp:
@iconofsin That looks good, but we would like to include a test with it - are you up for looking into it and adding one?
If not I can add a test to your branch.

iconofsin:
I cloned existing tests (based on `error One`) and added a decorated custom error in the contract to the best of my understanding.

rzadp (approved):

PR #798, opened 2022-09-26
👻 Add automatic ganache version upgrade workflow

jakvbs:


.github/workflows/upgrade-ganache.yaml
```diff
@@ -0,0 +1,31 @@
...
+          EXECUTE_COMMANDS: |
+            npm install -g npm-check-updates
+            cd packages/waffle-provider
+            ncu ganache -u
+            cd ..
+            cd waffle-ens/
+            ncu ganache -u
+          COMMIT_MESSAGE: "Upgrade ganache version"
+          COMMIT_NAME: "GitHub Actions"
+          COMMIT_EMAIL: "ci@example.com"
```
rzadp, line 28: Make it same as we have in changesets

.github/workflows/upgrade-ganache.yaml
```diff
@@ -0,0 +1,31 @@
...
+          PR_BRANCH_NAME: "upgrade-ganache-${PR_ID}"
+          PR_TITLE: "Upgrade ganache version"
+          GITHUB_TOKEN: ${{ secrets.CREATE_PR_TOKEN }}
```
rzadp, line 31:
```suggestion
          GITHUB_TOKEN: ${{ secrets. GH_BOT_TOKEN }}
```

.github/workflows/upgrade-ganache.yaml
```diff
@@ -0,0 +1,31 @@
...
+    timeout-minutes: 15
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 16.1
+      - name: Upgrade ganache version
+        uses: technote-space/create-pr-action@v2
+        with:
+          EXECUTE_COMMANDS: |
+            npm install -g npm-check-updates
```
rzadp, line 20: Pin the version - we don't want it to fail when breaking changes are released.

rzadp (approved):

PR #802, opened 2022-10-03
🗾 Extend matching of Hardhat revert reasons

rzadp:
The problem - usage of quotes or apostrophes in the revert strings.


yivlad (approved):

yivlad (approved):

yivlad (approved):

waffle-chai/src/matchers/revertedWith.ts
```diff
@@ -60,6 +61,13 @@ const decodeHardhatError = (error: any, context: any) => {
       context.txErrorName = error.errorName;
       return error.errorName;
     }
+    if (error?.code) {
```
yivlad, line 64: Shouldn't this be `error?.data`?

yivlad (approved):

yivlad (approved):

PR #808, opened 2022-10-31
🧇 Mark Waffle 4 as latest

rzadp:
Also remove internal optimism testing package from publihsing

### DIFF
--- a/waffle-optimism/package.json
+++ b/waffle-optimism/package.json
@@ -4,12 +4,8 @@
   "version": "4.0.2",
   "author": "Marek Kirejczyk <account@ethworks.io> (http://ethworks.io)",
   "repository": "git@github.com:EthWorks/Waffle.git",
-  "private": false,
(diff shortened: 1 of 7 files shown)


waffle-optimism/package.json
```diff
@@ -4,12 +4,8 @@
   "version": "4.0.2",
   "author": "Marek Kirejczyk <account@ethworks.io> (http://ethworks.io)",
   "repository": "git@github.com:EthWorks/Waffle.git",
-  "private": false,
+  "private": true,
   "license": "MIT",
-  "publishConfig": {
```
yivlad, line 9: Why? The package exports `OptimismProvider`

yivlad (approved):

waffle-optimism/package.json
rzadp: Rolledback

PR #810, opened 2022-11-08
🐦 Test hardhat-waffle-plugin submodule

rzadp:


.github/workflows/CI-plugin.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches:
+      - master
+jobs:
+  test-waffle-hardhat-plugin:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:        
+        hardhat: ['2.9.2', '2.11.2', 'latest']
+        hardhat-ethers: ['2.0.3', '2.1.1', 'latest']
+        hardhat-waffle: ['linked', 'latest'] # TODO: add dev (remember it's a different package)
```
rzadp, line 14:
```suggestion
        hardhat-waffle: ['linked', 'latest']
```

.github/workflows/CI-plugin.yml
```diff
@@ -0,0 +1,43 @@
+name: Plugin CI
```
rzadp, line 1: Remove those cases from `CI.yml`

waffle-chai/src/inject-call-history.ts
```diff
@@ -29,10 +29,18 @@ function toRecordedCall(message: any): RecordedCall {
...
 const inject = () => {
   let waffle: any;
+  let config: any;
   try {
-    waffle = require('hardhat')?.waffle;
+    ({waffle, config} = require('hardhat'));
   } catch { return; }
   if (!waffle || !waffle.provider) return;
+  if (!config) {
+    console.warn('Missing hardhat config, skipping call history injection');
```
rzadp, line 38:
```suggestion
    console.warn('Hardhat config not present, skipping call history injection');
```

Use lighter wording - "missing" suggests it's a mistake and the user should add it - but if the user uses Ganache then it is not a mistake.

yivlad (approved):

PR #815, opened 2023-01-09
🗽 Add address parameter to the mock contract

pawelpolak2:


waffle-mock-contract/test/directTest.ts
```diff
@@ -0,0 +1,132 @@
...
+      const address = Wallet.createRandom().address;
+      const counterFactory = new ContractFactory(Counter.abi, Counter.bytecode, wallet);
+      const counter = await counterFactory.deploy();
+      const mockCounter = await deployMockContract(wallet, Counter.abi, address);
+
+      await mockCounter.call(counter, 'increaseByThreeValues', 1, 2, 3);
+      expect(await counter.read()).to.equal('6');
+    });
+
+    it('can be deployed under specified address', async () => {
```
yivlad, line 116: I think it's a good idea to also have the intended flow, meaning that we have hardcoded address in some contract and we want to mock the contract under this hardcoded address. Then the first contract calls the mocked contract and gets some result, using it further in it's computations (can be as simple as `require(result == 1)`)

waffle-mock-contract/src/index.ts
```diff
@@ -15,7 +16,28 @@ export interface MockContract extends Contract {
   staticcall (contract: Contract, functionName: string, ...params: any[]): Promise<any>;
 }
 
-async function deploy(signer: Signer) {
+async function deploy(signer: Signer, address?: string) {
```
yivlad, line 19: Consider adding an option allowing to override the existing address. So the signature will look something like 
```ts
type DeployOptions = {
  address?: string
} | {
  address: string,
  overrideExsiting: boolean
}

async function deploy(signer: Signer, options?: DeployOptions = {}) {
```

yivlad (changes requested):

waffle-mock-contract/test/proxiedTest.ts
```diff
@@ -0,0 +1,61 @@
...
+    it('mocking with call arguments', async () => {
+      const {capContract, mockCounter} = await deploy();
+      await mockCounter.mock.add.withArgs(1).returns(1);
+      await mockCounter.mock.add.withArgs(2).returns(2);
+
+      expect(await capContract.addCapped(1)).to.equal(1);
+      expect(await capContract.addCapped(2)).to.equal(2);
+    });
+
+    it('specified address intended usage', async () => {
```
yivlad, line 52: Try to figure out a better name - describing what's inside the test

docs/source/mock-contract.rst
```diff
@@ -16,6 +16,11 @@ Create an instance of a mock contract providing the :code:`ABI` of the smart con
 
   const mockContract = await deployMockContract(wallet, contractAbi);
 
+You can also choose the deployment address of the mock contract by providing it as an argument:
```
yivlad, line 19: Update docs according to the latest changes

yivlad (approved):

PR #816, opened 2023-01-16
🖼 Mock contract chaining behaviour

pawelpolak2:


docs/source/mock-contract.rst
```diff
@@ -123,6 +123,55 @@ Mock contract will be used to mock exactly this call with values that are releva
     });
   });
 
+Mocking multiple calls
+----------------------
+
+Mock contract allows to queue multiple mock calls to the same function. This can only be done if the function is not pure or view. That's because the mock call queue is stored on the blockchain and we need to modify it.
```
yivlad, line 129: Write a test where we call view function during a transaction. For instance, reading how many tokens do we own, then somehow update this number, then read again. Change the state accordingly. Also update the docs.

waffle-mock-contract/src/Doppelganger.sol
```diff
@@ -61,15 +89,25 @@ contract Doppelganger {
       return returnValue;
     }
 
-    function __internal__getMockCall() view private returns (MockCall storage mockCall) {
-        mockCall = mockConfig[keccak256(msg.data)];
-        if (mockCall.initialized == true) {
+    function __internal__getMockCall() private returns (MockCall memory mockCall) {
+        bytes32 root = keccak256(msg.data);
+        mockCall = mockConfig[root];
+        if (mockCall.next != "") {
```
yivlad, line 95: Add comments explaining what is going on

waffle-mock-contract/src/Doppelganger.sol
```diff
@@ -3,45 +3,73 @@ pragma solidity ^0.6.3;
...
     struct MockCall {
-        bool initialized;
+        bytes32 next;
         bool reverts;
         string revertReason;
         bytes returnValue;
     }
 
     mapping(bytes32 => MockCall) mockConfig;
+    mapping(bytes32 => bytes32) heads;
```
yivlad, line 13: Rename

waffle-mock-contract/src/index.ts
```diff
@@ -6,21 +6,120 @@ import type {JsonRpcProvider} from '@ethersproject/providers';
...
+    this.revertSet = false;
+    this.argsSet = false;
+    throw new Error(reason);
+  }
+
+  returns(...args: any) {
+    if (this.revertSet) this.err('Revert must be the last call');
+    if (!this.func.outputs) return this; // Throw ?
+    const encoded = this.encoder.encode(this.func.outputs, args);
+    if (this.stubCalls.length === 0) {
```
yivlad, line 49: Add comments explaining what is going on

yivlad (changes requested):

docs/source/mock-contract.rst
pawelpolak2: Seems like this is not possible, because calling view functions uses a STATICCALL opcode.

yivlad (approved):

Issue #809, opened 2022-11-06
[Feature Request] Emit events in mock contracts


sthitpragnya-ethosx:
Is there a way we can specify a mock contract to emit an event when a write function is called? This would come in handy to check whether a particular functions is called or not. A good to have would be to also map the parameters of the event to input params to test the parameters with which the mock's function is called.

rzadp:
@sthitpragnya-ethosx Thanks for submitting the request.

> Is there a way we can specify a mock contract to emit an event when a write function is called?

No, not at the moment. Let's scope out what would be the use cases for it.

> This would come in handy to check whether a particular functions is called or not.
> (...) to test the parameters with which the mock's function is called.

So, it seems like what you would need is not necessarily events, but a way to assert `calledOnContract` and `calledOnContractWith` on the mock contract, right?

pawelpolak2:
`calledOnContract` and `calledOnContractWith` matchers work correctly with mock contracts which seems to resolve this.

PR #822, opened 2023-02-01
⭐️ Support events not defined in a contract

### LINKED ISSUE #594: .to.emit does not work with delegatecall
## Description

Attempting to test for emission of events using chai matchers on a contract that delegates its call to the source of the event does not find the emitted event.

```javascript
await caller.deployed();
await target.deployed();
...
await expect(caller.emitEvent()).to.emit(target, 'Test');
```

Here are slimmed versions of the contracts I used.

```solidity
contract Caller {
    function emitEvent(address target) public {
        bytes4 funcSig = bytes4(keccak256(abi.encodePacked("emitEvent()")));
        address implementation = address(target);
        assembly {
            success := delegatecall(gas(), implementation, funcSig, 4, 0, 0)
            returndatacopy(0, 0, returndatasize())

            switch success
            case 0 {
                revert(0, returndatasize())
            }
            default {
                return(0, returndatasize())
            }
        }
    }
}
```

```solidity
contract Target {
    event Test();
    function emitEvent() public {
        emit Test();
    }
}
```

`Caller` contract has a function `emitEvent` that performs a delegate call to a similar function in `Target` contract that emits an event `Test`. Since the delegate call performs all state changes from the context of the calling contract `Caller`, the event is emitted from `Caller` even though it is defined by `Target`.

Using the syntax `await expect(caller.emitEvent()).to.emit(target, 'Test');` results in:

```bash
AssertionError: Expected event "Test" to be emitted, but it wasn't
```

even though receipt logs show:
```javascript
logs: [
    {
      transactionIndex: 0,
      blockNumber: 5,
      transactionHash: '0x66978ebc21c3126c307afbb7f477827bb2cd5c12b26f8185ac9d99b5dbdf1c13',
      address: '0x5FbDB2315678afecb367f032d93F642f64180aa3',
      topics: [
        '0xa163a6249e860c278ef4049759a7f7c7e8c141d30fd634fda9b5a6a95d111a30' 
      ],
      data: '0x',
      logIndex: 0,
      blockHash: '0x00565d9199d04b6b4ae1a2c4eb1b3b869c040e1b1234d5156337ef32dfd2b04d'
    }
  ]
```
where the topics array contains the `keccak256("Test()") == "0xa163a6249e860c278ef4049759a7f7c7e8c141d30fd634fda9b5a6a95d111a30"`

This leads me to believe that the syntax used causes waffle to expect the event to be emitted from `Target` but was in fact emitted by `Caller` and attempting to change the emission clause to `.to.emit(caller, 'Test');` gives us:

```bash
AssertionError: Expected event "Test" to be emitted, but it doesn't exist in the contract. Please make sure you've compiled its latest version before running the test.
      + expected - actual

      +Test

```
as it attempts to find the event definition in `Caller` but it only exists in `Target`.

## Versions
macOS 11.3.1

``` javascript
{
  'core-contracts': '1.0.0',
  npm: '7.20.3',
  node: '16.6.2',
  v8: '9.2.230.21-node.19',
  uv: '1.41.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.17.2',
  modules: '93',
  nghttp2: '1.42.0',
  napi: '8',
  llhttp: '6.0.2',
  openssl: '1.1.1k+quic',
  cldr: '39.0',
  icu: '69.1',
  tz: '2021a',
  unicode: '13.0',
  ngtcp2: '0.1.0-DEV',
  nghttp3: '0.1.0-DEV'
}
```

```json
"devDependencies": {
    "@nomiclabs/hardhat-ethers": "^2.0.2",
    "@nomiclabs/hardhat-waffle": "^2.0.1",
    "@nomiclabs/hardhat-web3": "^2.0.0",
    "chai": "^4.3.4",
    "ethereum-waffle": "^3.4.0",
    "ethers": "^5.4.4",
    "hardhat": "^2.6.0",
    "web3": "^1.5.2"
  }
```

pawelpolak2:
closes #594


waffle-chai/src/matchers/emit.ts
```diff
@@ -1,16 +1,30 @@
...
     logs.filter((log) => log.topics.includes(topic))
-      .filter((log) => log.address && log.address.toLowerCase() === contractAddress.toLowerCase());
+      .filter((log) =>
+        log.address &&
+        (contractAddress === undefined || log.address.toLowerCase() === contractAddress.toLowerCase()
+        ));
 
-  Assertion.addMethod('emit', function (this: any, contract: Contract, eventName: string) {
+  const getEventName = (eventSig: string) => {
+    const reg = /(\S+)\((\S*)\)/;
```
yivlad, line 17: Isn't there a way to parse event signature using some method form ethers?

waffle-chai/src/matchers/emit.ts
```diff
@@ -1,16 +1,30 @@
...
+    const matches = reg.exec(eventSig);
+    if (matches) return matches[1];
+    else throw new Error(`Invalid event signature "${eventSig}"`);
+  };
+
+  Assertion.addMethod('emit', function (this: any, contract: Contract|string, eventName?: string) {
```
yivlad, line 23: Don't like this. If necessary consider defining function in a separate place and give it to possible signatures. Also it's better to use self explanatory variable names - e.g. not contract but contractOrEventName

waffle-chai/test/matchers/eventsTest.ts
```diff
@@ -698,5 +699,43 @@ export const eventsWithNamedArgs = (provider: TestProvider) => {
         '{ Object (hash, value, ...) } to deeply equal { Object (hash, value, ...) }'
       );
     });
+
+    it('Signature only - delegatecall', async () => {
+      const proxyFactory = new ContractFactory(EventsProxyABI.abi, EventsProxyABI.bytecode, wallet);
+      const proxy = await proxyFactory.deploy(events.address);
+
+      await expect(proxy.emitTwoDelegate()).to.emit('Two(uint256,string)');
```
yivlad, line 707: There is one more scenario - transaction on contract A calls a function on contract B and contract B emits an event. We want to be able to capture this event as well.

yivlad (changes requested):

yivlad:
Also changest

waffle-chai/package.json
```diff
@@ -42,6 +42,7 @@
   },
   "dependencies": {
     "@ethereum-waffle/provider": "workspace:*",
+    "chai": "4.2.0",
```
yivlad, line 45: Why?

yivlad: Also why so many changes to lockfile? I don't think they're necessary

waffle-chai/src/matchers/emit.ts
```diff
@@ -1,70 +1,88 @@
...
+      if (typeof contractOrEventSig === 'string') {
+        try {
+          eventFragment = utils.EventFragment.from(contractOrEventSig);
+        } catch (e) {
+          throw new Error(`Invalid event signature: "${contractOrEventSig}"`);
         }
-        const receipt: providers.TransactionReceipt = this.txReceipt;
-        let eventFragment: utils.EventFragment | undefined;
+        assertEmit(this, eventFragment, isNegated);
+      } else if (typeof contractOrEventSig === 'object' && eventName) {
```
yivlad, line 55: `typeof contractOrEventSig === 'object'` - is it really necessary? I think typescript type system should do this for us

waffle-mock-contract/test/proxiedTest.ts
```diff
@@ -57,5 +57,21 @@ export function mockContractProxiedTest(provider: MockProvider) {
       await mockContract.mock.read.returns(1);
       expect(await proxy.readCapped()).to.eq(1);
     });
+
+    it('calledOnContract with mock contract', async () => {
```
yivlad, line 61: What's the reason for adding these tests?

yivlad (changes requested):

waffle-mock-contract/test/proxiedTest.ts
pawelpolak2: I wrote these for #809 and forgot about it.

yivlad (approved):

Missjones2829 (commented):
Merge

